PR #15, opened 2024-02-01
CDMD-2459 [React Router Codemod] Replace location.query

hbjORbj:
-   **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

- Adds one codemod for migrating React Router from v3 to v4. This codemod replaces instances of `location.query` with `parse(location.search)`, where `parse` is a function imported from `query-string`.

### DIFF
--- a/apps/registry/codemods/react-router/4/replace-location-query/.mocharc.json
+++ b/apps/registry/codemods/react-router/4/replace-location-query/.mocharc.json
@@ -1,0 +1,8 @@
+{
+	"loader": ["ts-node/esm"],
+	"full-trace": true,
+	"failZero": false,
+	"bail": true,
+	"spec": "./**/test.ts",
+	"timeout": 5000
+}
--- a/apps/registry/codemods/react-router/4/replace-location-query/.mocharc.json
+++ b/apps/registry/codemods/react-router/4/replace-location-query/.mocharc.json
@@ -1,8 +1,0 @@
-{
-	"loader": ["ts-node/esm"],
-	"full-trace": true,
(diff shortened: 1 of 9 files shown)


apps/registry/codemods/react-router/4/replace-location-query/.mocharc.json
```diff
@@ -0,0 +1,8 @@
+{
```
DmytroHryshyn, line 1: do we need this file?

apps/registry/codemods/react-router/4/replace-location-query/src/index.ts
```diff
@@ -0,0 +1,44 @@
...
+	);
+
+	const hasQueryStringImport =
+		root.find(j.ImportDeclaration, {
+			source: {
+				value: 'query-string',
+			},
+		}).length > 0;
+
+	if (!hasQueryStringImport) {
```
DmytroHryshyn, line 30: will this codemod insert `query-string` import even if we did not replace `location.query`? I think we need to add a `dirtyFlag` and insert import only when we replaced location.query

hbjORbj: addressed and added a test

apps/registry/codemods/react-router/4/replace-location-query/.mocharc.json
hbjORbj: nope, addressed

DmytroHryshyn (dismissed):

PR #19, opened 2024-02-01
fix: Make build:cjs run first and add build action for PR

r4zendev:


grzpab (dismissed):

.github/workflows/build.yml
```diff
@@ -12,15 +12,16 @@ jobs:
               with:
                   node-version: 20
             - name: Checkout Repository
-              uses: actions/checkout@v2
+              uses: actions/checkout@v3
```
grzpab, line 15: out of curiosity, why was it needed?

.github/workflows/build.yml
```diff
@@ -12,15 +12,16 @@ jobs:
...
                   token: ${{ secrets.GH_TOKEN }}
             - name: Install modules
               uses: pnpm/action-setup@v2
               with:
                   version: 8
-            - name: Install dependencies
-              run: pnpm install
+                  run_install: |
+                      - recursive: true
+                        args: [--frozen-lockfile, --strict-peer-dependencies]
```
grzpab, line 24: I guess the peer deps were the problem?

grzpab (commented):
LGTM, let's merge it tomorrow morning and test it

.github/workflows/build.yml
r4zendev: i think recursive was the solution and adding cjs-builder to deps of registry also helped.

.github/workflows/build.yml
r4zendev: for funzies, no specific reason, just to stay with the latest version

grzpab (approved):

PR #29, opened 2024-02-06
chore: implement CLI banner, replace codemod engine commands

DmytroHryshyn:
https://github.com/codemod-com/codemod/assets/125881252/f57e6614-8779-45a6-a387-9972784b121c

https://linear.app/codemod/issue/CDMD-2485/implement-error-banner-ui-that-is-displayed-when
https://linear.app/codemod/issue/CDMD-2486/replace-executable-calls-with-global-cli-calls-remove-code-that-is


apps/vsce/webview/src/CodemodEngineNodeNotFound/index.tsx
```diff
@@ -0,0 +1,47 @@
+import { VSCodeTextField } from '@vscode/webview-ui-toolkit/react';
+import cn from 'classnames';
+import styles from './style.module.css';
+
+const INSTALL_CODEMOD_ENGINE_NODE_COMMAND = 'npm install --global codemod';
```
grzpab, line 5: we might shorten it into `npm i -g codemod`

apps/vsce/src/components/engineService.ts
```diff
@@ -228,30 +230,43 @@ export class EngineService {
...
+					kind: MessageKind.codemodEngineNodeLocated,
+				});
+				this.__onCodemodEngineNodeLocated();
+				clearInterval(codemodEnginePollingIntervalId);
+			}
+		};
+
+		checkCodemodEngineNode();
+		// we retry codemod engine installation checks automatically, so we can detect when user installs the codemod
+		const codemodEnginePollingIntervalId = setInterval(
```
grzpab, line 253: when is this interval cancelled?

DmytroHryshyn: Its cancelled when we successfully locate the engine, in the checkCodemodEngineNode function

grzpab: Wow, I would assume TS would comment on such an usage of the interval, can we constraint it as well to the max number of checks (e.g. 10?)

grzpab (approved):

PR #37, opened 2024-02-09
fix(mocha-vitest): create vitest.config.ts file on codemod run

r4zendev:


apps/registry/codemods/mocha/vitest/migrate-configuration/src/index.ts
```diff
@@ -8,6 +8,7 @@ import {
 	string,
 	type Input,
 } from 'valibot';
+import { FileCommand } from '../../../../../../../packages/filemod/src/internalCommands.js';
```
grzpab, line 11: that's a type import right?

grzpab (approved):
looks good to me

PR #50, opened 2024-02-14
CDMD-2542 Improve displayed message for codemod list

hbjORbj:


r4zendev:
do you think it's needed? if you're looking to improve the codemod list command i would look for a way to display it in multiple rows and make them colored based on if the codemod is verified or not, if we have access to it (later, we ideally should have it, to tell the user what are the codemods that are safe to run)

what i mean by rows can be seen when using `lsd` CLI:
image

hbjORbj:
yes, I absolutely agree with you on this. I will close the PR.

PR #53, opened 2024-02-14
fix: make prettier work for most file types CDMD-2529

r4zendev:
- Downgrades prettier to v2 for CLI to make it bundle successfully with esbuild, which for some reason does not work properly for v3.
- Adds a mapping for most used file extensions to support formatting them and not only css/ts files.
- Adjusts lint-staged rule to include all the files that we want to lint/format (in particular, packages/**)


DmytroHryshyn (approved):

hbjORbj (dismissed):

mohab-sameh:
@all-contributors add @r4zendev for code and bug fixes

PR #66, opened 2024-02-20
Add ast-grep as an engine 

Harsh24893:
# This Diff 
**Summary**
- In commit 1, adding no-op ast-grep engine to codemod cli 
- In commit 2, added implementation of ast-grep engine to cli 

**Testing**

I put the following code in a file called abc.js in folder `/Users/harshgupta/Desktop/src/codemod/apps/cli/ast/abc.js`
```
import { A, B, C } from './barrel';

function Component() {
    const [name, setName] = useState < string > ('React')
}
``` 

In directory `/Users/harshgupta/.codemod`, chose a random folder `XG6X5ceI2dzkLW7CoBI0aD3W_7Q`, and updated it's config.json to have engine `ast-grep`. 

config.json after update 
```
"schemaVersion":"1.0.0","engine":"ast-grep","arguments":[],"name":"cal.com/app-directory-boilerplate-calcom"}
```

Also added a file called rule.yaml in `XG6X5ceI2dzkLW7CoBI0aD3W_7Q` and content of yaml file was 
```
id: change_js
language: javascript
rule:
  pattern: import {$$$IDENTS} from './barrel'
rewriters:
- id: rewrite-identifer
  rule:
    pattern: $IDENT
    kind: identifier
  fix: import $IDENT from './barrel/$IDENT'
transform:
  IMPORTS:
    rewrite:
      rewriters: [rewrite-identifer]
      source: $$$IDENTS
      joinBy: "\n"
fix: $IMPORTS
```

Built the codemod using 
```npm run build && npm link```

Executed 
```codemod cal.com/app-directory-boilerplate-calcom``` with useCache true and --target=/Users/harshgupta/Desktop/src/codemod/apps/cli/ast 

**useCache set to true so that it doesn't redownload the codemod from S3 which would update config.json and overwrite the engine.

**Resources**
https://ast-grep.github.io/guide/introduction.html 

# Follow up 
- Come up with some codemod for ast-grep and publish to registry 
- Add a script to execute ast-grep codemods in a batch manner and create PRs from it

### DIFF
--- a/apps/cli/src/downloadCodemod.ts
+++ b/apps/cli/src/downloadCodemod.ts
@@ -124,10 +124,10 @@ export class CodemodDownloader implements CodemodDownloaderBlueprint {
 
 		if (config.engine === "ast-grep") {
 			try {
-				const yamlPath = join(directoryPath, 'rule.yaml');
+				const yamlPath = join(directoryPath, "rule.yaml");
 
 				return {
-					source: 'registry',
+					source: "registry",
 					name,
 					engine: config.engine,
 					yamlPath,
@@ -136,18 +136,18 @@ export class CodemodDownloader implements CodemodDownloaderBlueprint {
 				};
 			} catch (error) {
 				if (!(error instanceof Error)) {
-					throw new Error('Error while downloading ast-grep codemod');
+					throw new Error("Error while downloading ast-grep codemod");
 				}
 
 				this.__printer.printOperationMessage({
-					kind: 'error',
+					kind: "error",
 					message: error.message,
 				});
 			}
 		}
 
 		if (config.engine === "piranha") {
-			const rulesPath = join(directoryPath, 'rules.toml');
+			const rulesPath = join(directoryPath, "rules.toml");
 
 			await this._fileDownloadService.download(
 				`${CODEMOD_REGISTRY_URL}/${hashDigest}/rules.toml`,
--- a/apps/cli/src/downloadCodemod.ts
+++ b/apps/cli/src/downloadCodemod.ts
@@ -123,22 +123,27 @@ export class CodemodDownloader implements CodemodDownloaderBlueprint {
 		}
 
 		if (config.engine === "ast-grep") {
-			const yamlPath = join(directoryPath, 'rule.yaml');
-			const yamlData = await this._fileDownloadService.download(
-				`${CODEMOD_REGISTRY_URL}/${hashDigest}/rule.yaml`,
-				yamlPath,
-			);
+			try {
+				const yamlPath = join(directoryPath, 'rule.yaml');
 
-			await writeFile(yamlPath, yamlData);
+				return {
+					source: 'registry',
+					name,
+					engine: config.engine,
+					yamlPath,
+					directoryPath,
+					arguments: config.arguments,
+				};
+			} catch (error) {
+				if (!(error instanceof Error)) {
+					throw new Error('Error while downloading ast-grep codemod');
+				}
 
-			return {
-				source: 'registry',
-				name,
-				engine: config.engine,
-				yamlPath: yamlPath,
-				directoryPath,
-				arguments: config.arguments,
-			};
+				this.__printer.printOperationMessage({
+					kind: 'error',
+					message: error.message,
+				});
+			}
 		}
 
 		if (config.engine === "piranha") {
--- a/apps/cli/src/downloadCodemod.ts
+++ b/apps/cli/src/downloadCodemod.ts
@@ -122,8 +122,27 @@ export class CodemodDownloader implements CodemodDownloaderBlueprint {
 			// do nothing, descriptions might not exist
 		}
 
+		if (config.engine === "ast-grep") {
+			const yamlPath = join(directoryPath, 'rule.yaml');
+			const yamlData = await this._fileDownloadService.download(
+				`${CODEMOD_REGISTRY_URL}/${hashDigest}/rule.yaml`,
+				yamlPath,
+			);
+
+			await writeFile(yamlPath, yamlData);
+
+			return {
+				source: 'registry',
+				name,
+				engine: config.engine,
+				yamlPath: yamlPath,
+				directoryPath,
+				arguments: config.arguments,
+			};
+		}
+
 		if (config.engine === "piranha") {
-			const rulesPath = join(directoryPath, "rules.toml");
+			const rulesPath = join(directoryPath, 'rules.toml');
 
 			await this._fileDownloadService.download(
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by Harsh24893** [comment]
I am seeing some build failures as it is not able to figure out js-yaml files, but I don't see failures while executing codemod.

File: apps/cli/src/downloadCodemod.ts [hunk 1]
```diff
@@ -111,6 +111,25 @@ export class CodemodDownloader implements CodemodDownloaderBlueprint {
...
+				yamlPath,
+			);
+
+			await writeFile(yamlPath, yamlData);
+
+			return {
+				source: 'registry',
+				name,
+				engine: config.engine,
+				yamlPath: yamlPath,
```
r4zendev, line 127: use shorthand

apps/cli/src/runAstgrepCodemod.ts
```diff
@@ -0,0 +1,62 @@
...
+				// Recursively call the function for subdirectories
+				await iterateFiles(entryPath);
+			} else if (entry.isFile()) {
+				// If the entry is a file, log its path
+				const fileExtension = path.extname(entryPath).slice(1);
+				if (fileExtension !== language) {
+					continue;
+				}
+				const astCommand = `sg scan -r ${rulePath} ${entryPath} -U`;
+				execSync(astCommand);
```
r4zendev, line 41: i'm not sure this is going to work unless user has installed ast-grep on his machine

apps/cli/src/downloadCodemod.ts
```diff
@@ -111,6 +111,25 @@ export class CodemodDownloader implements CodemodDownloaderBlueprint {
 			}
 		}
 
+		if (config.engine === 'ast-grep') {
+			const yamlPath = join(directoryPath, 'rule.yaml');
+			const yamlData = await this._fileDownloadService.download(
```
DmytroHryshyn, line 116: can it throw? Should we wrap it in try/catch?

apps/cli/src/downloadCodemod.ts
```diff
@@ -111,6 +111,25 @@ export class CodemodDownloader implements CodemodDownloaderBlueprint {
...
+				`${CODEMOD_REGISTRY_URL}/${hashDigest}/rule.yaml`,
+				yamlPath,
+			);
+
+			await writeFile(yamlPath, yamlData);
```
DmytroHryshyn, line 121: I think `download` method already saves file to the filesystem

apps/cli/src/runAstgrepCodemod.ts
```diff
@@ -0,0 +1,62 @@
...
+	const yamlObject = yaml.load(yamlString);
+	const language = languageToExtension(yamlObject.language);
+	printer.printConsoleMessage(
+		'info',
+		`Executing ast-grep for language : ${language}`,
+	);
+
+	// Function to recursively iterate over files in the directory
+	const iterateFiles = async (dirPath: string) => {
+		const entries = await fs.promises.readdir(dirPath, {
```
DmytroHryshyn, line 23: do we need to take into account `include` and `exclude` patterns here (passed as cli args)?

DmytroHryshyn: maybe we can use `glob` lib for this use case?

apps/cli/src/runAstgrepCodemod.ts
```diff
@@ -0,0 +1,62 @@
...
+			if (entry.isDirectory()) {
+				// Recursively call the function for subdirectories
+				await iterateFiles(entryPath);
+			} else if (entry.isFile()) {
+				// If the entry is a file, log its path
+				const fileExtension = path.extname(entryPath).slice(1);
+				if (fileExtension !== language) {
+					continue;
+				}
+				const astCommand = `sg scan -r ${rulePath} ${entryPath} -U`;
```
DmytroHryshyn, line 40: will this work on `Windows`? i think we need to add quotes to the args to make it work in Windows cmd. I will check later.

apps/cli/src/runAstgrepCodemod.ts
r4zendev: also we don't need to execute changes here, we need to build new files, then execute buildFileCommands() and pass it up. similar how it's done for other engines. not sure if you're doing the same here.

apps/cli/src/runAstgrepCodemod.ts
```diff
@@ -0,0 +1,62 @@
...
+import { PrinterBlueprint } from './printer.js';
+
+export const runAstgrep = async (
+	printer: PrinterBlueprint,
+	rulePath: string,
+	targetDirectory: string,
+): Promise<void> => {
+	const yamlString = await readFile(rulePath, { encoding: 'utf8' });
+	const yamlObject = yaml.load(yamlString);
+	const language = languageToExtension(yamlObject.language);
```
DmytroHryshyn, line 15: lets call this variable `extension`

mohab-sameh:
@all-contributors add @Harsh24893 for code

apps/cli/src/downloadCodemod.ts
Harsh24893: It shouldn't throwing, but adding a try catch is always a good idea. Will do it.

apps/cli/src/runAstgrepCodemod.ts
Harsh24893: I was thinking of shipping this and doing other changes once Sasha refactors the cli.

apps/cli/src/runAstgrepCodemod.ts
Harsh24893: Updated the code

apps/cli/src/runAstgrepCodemod.ts
Harsh24893: I think just installing it for the user is a good idea to get ahead. Also this ll be in beta mode, so we can work on better devX as we move ahead. 

What do you mean by build new files.

apps/cli/src/runAstgrepCodemod.ts
```diff
@@ -0,0 +1,87 @@
...
+// Function to check if ast-grep is available or not
+const installSgCommandIfNotAvailable = (printer: PrinterBlueprint): void => {
+	try {
+		// Use `which` command to check if the command is available
+		execSync(`which sg`);
+	} catch (error) {
+		// If `which` command fails, the command is not available
+		printer.printConsoleMessage(
+			'info',
+			'ast-grep is not avialable, installing it globally',
```
r4zendev, line 78: available

r4zendev (dismissed):

r4zendev (approved):

PR #69, opened 2024-02-21
CDMD-2542 Improve UI/UX for codemod list

hbjORbj:
- ensure `owner` field in config.json is not neglected
- improve UI/UX of codemod list with 3 columns (name, engine, owner)
- color and make bold the text if codemods are written by `codemod.com` team (we currently do not have 'verified' field in config.json. we could do this in the future if needed)


apps/cli/src/handleListCliCommand.ts
```diff
@@ -22,26 +24,71 @@ export const handleListNamesCommand = async (printer: Printer) => {
...
 		configFiles.map(async (cfg) => {
 			const configJson = await readFile(cfg, 'utf8');
 
 			const parsedConfig = v.safeParse(
-				v.object({ name: v.string() }),
+				v.object({
+					name: v.string(),
+					engine: v.string(),
+					owner: v.optional(v.string()),
+				}),
```
r4zendev, line 36: i would make this schema a const somewhere else, since it's reused below L71

r4zendev: Also, I'd make owner optional as well. We can handle this on our side, and it's better if we don't make it strict. Do you think it will always be present? What about download feature? What about `pnpm create` command in registry that creates a boilerplate? Not sure

r4zendev (approved):

PR #80, opened 2024-02-22
feat: execute list command on codemod not found (CDMD-2590)

r4zendev:
- Does not show that codemod is executing before actual download happens.

image

- Now instead of showing 403 request error from axios, we show clear error saying that codemod wasn't found and show user a list of commands.

image

### DIFF
--- a/apps/cli/src/downloadCodemod.ts
+++ b/apps/cli/src/downloadCodemod.ts
@@ -106,7 +106,9 @@ export class CodemodDownloader implements CodemodDownloaderBlueprint {
 			await handleListNamesCommand(this.__printer);
 
 			throw new Error(
-				`Could not find codemod ${name} in the registry. Verify the name to be in the list above and try again.`,
+				`Could not find codemod ${boldText(
+					name,
+				)} in the registry. Verify the name to be in the list above and try again.`,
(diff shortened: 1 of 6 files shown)


apps/cli/src/downloadCodemod.ts
```diff
@@ -86,29 +90,42 @@ export class CodemodDownloader implements CodemodDownloaderBlueprint {
...
+			);
 
-		const config = S.parseSync(codemodConfigSchema)(parsedConfig);
+			parsedConfig = JSON.parse(buffer.toString('utf8'));
+		} catch (error) {
+			await handleListNamesCommand(this.__printer);
 
-		{
-			const descriptionPath = join(directoryPath, 'description.md');
+			throw new Error(
```
DmytroHryshyn, line 108: should we check if error is actually axios `404` error before throwing this?

DmytroHryshyn (approved):

apps/cli/src/downloadCodemod.ts
r4zendev: it's actually 403 because of how AWS handles this, but i could add a check. not sure if it's needed though

PR #85, opened 2024-02-23
feat: React Compiler useContext codemod (CDMD-2578)

r4zendev:
This PR adds a codemod that would migrate context usages in React v19 to using `use` hook.

### DIFF
--- a/apps/registry/codemods/react/19/use-context-hook/test/test.ts
+++ b/apps/registry/codemods/react/19/use-context-hook/test/test.ts
@@ -1,0 +1,6 @@
+// import assert from 'node:assert';
+// import { buildApi } from '@codemod-com/utilities';
+// import { describe, it } from 'vitest';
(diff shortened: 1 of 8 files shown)


apps/registry/codemods/react/19/use-context-hook/test/test.ts
```diff
@@ -0,0 +1,37 @@
+import assert from "node:assert/strict";
+import { buildApi } from "@codemod-com/utilities";
+import type { FileInfo } from "jscodeshift";
+import { describe, it } from "vitest";
+import transform from "../src/index.js";
+
+describe("useContext -> use", () => {
+	it("should replace useContext with use", async () => {
+		const input = `
+    import { useContext } from "react";
```
hbjORbj, line 10: does your codemod handle `React.useContext` as well?

hbjORbj (approved):

PR #86, opened 2024-02-23
fix: fix vsce empty registry

DmytroHryshyn:
fix emtpy registry issue

### DIFF
--- a/apps/cli/src/executeMainThread.ts
+++ b/apps/cli/src/executeMainThread.ts
@@ -9,9 +9,9 @@ import yargs from "yargs";
 import { hideBin } from "yargs/helpers";
 import { buildArgumentRecord } from "./buildArgumentRecord.js";
 import {
-	buildOptions,
-	buildUseCacheOption,
-	buildUseJsonOption,
+  buildOptions,
+  buildUseCacheOption,
+  buildUseJsonOption,
 } from "./buildOptions.js";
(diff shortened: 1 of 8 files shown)


apps/cli/src/executeMainThread.ts
```diff
@@ -60,6 +62,13 @@ export const executeMainThread = async () => {
 
 	if (!process.stdin.isTTY) {
 		await new Promise((resolve) => {
+			setTimeout(() => {
+				if (userInput.trim() === '') {
+					// skip if no input in 1000 ms
+					resolve(null);
+				}
+			}, WAIT_INPUT_TIMEOUT);
+
```
r4zendev, line 71: this is exactly what i was trying to avoid. i wanted to find a way to make it either read if there is input and if not, just go with the usual flow. instead, we are making our CLI slower on purpose. i don't really like this approach, but we can leave it like that for the time being and create a task to find a better way. what do you think? also, can you try at least lowering it to 500 ms? does it work?

DmytroHryshyn: agree, i just did the most straightforward fix and i dont like it. Will taskify so we can fix it later. Will set timeout to 300, should work for now

r4zendev (dismissed):

r4zendev (approved):

PR #91, opened 2024-02-23
cdmd-2575 [react codemod] remove-memoization-hooks

hbjORbj:


apps/registry/codemods/react/19/remove-memoization-hooks/src/index.ts
```diff
@@ -0,0 +1,46 @@
...
+				}
+				return specifier;
+			}) ?? [];
+
+		if (specifiers.length === 0) {
+			j(path).remove();
+		} else {
+			path.node.specifiers = specifiers;
+		}
+	});
```
r4zendev, line 26: you are removing each import that has no import specifiers. this would remove side-effect imports, such as `import "file.css"`
i would try removing this part, as you are removing identifiers below. you probably don't even need to clear the import statements.

hbjORbj: > you are removing each import that has no import specifiers. this would remove side-effect imports, such as import "file.css"

Good point! I pushed a fix.

PR #126, opened 2024-03-05
[CLI] fix: new validation for codemod configuration (CDMD-2655)

r4zendev:


apps/registry/builder/index.ts
```diff
@@ -63,41 +76,101 @@ export const PIRANHA_LANGUAGES = [
...
+];
+
+const versionValidator = union([
+	// react < 18.0.2 (preferred)
+	string([regex(semVerRegex)]),
+	// react < 18 (for example, when no latest version of a given major is out yet)
+	coerce(number(), (input) => Number(input)),
+]);
+
+const configJsonBaseSchema = object({
```
DmytroHryshyn, line 98: can we reuse `configJsonBaseSchema`? currently its duplicated in cli and registry

r4zendev: To be honest, in my opinion, this builder functionality is going to be partly moved to our backend, where we will re-upload registry.tar.gz every time a new public codemod is pushed from the CLI, and registry will not be responsible for uploading files to s3 anymore. This builder is subject to deprecation, so I think no need to find a solution for that issue right now.

DmytroHryshyn (approved):

PR #145, opened 2024-03-08
[Registry] [New Codemod] Deprecate nested routes - CDMD-2755

hbjORbj:


packages/codemods/react-router/4/replace-nested-routes/src/index.ts
```diff
@@ -0,0 +1,80 @@
...
+	const j = api.jscodeshift;
+	const root = j(file.source);
+
+	const routes = root.find(j.JSXElement, {
+		openingElement: {
+			name: {
+				name: "Route",
+			},
+		},
+		closingElement: {
```
DmytroHryshyn, line 13: what if Route is selfclosing jsx tag?

hbjORbj: We are looking for Route components that nest Route components

DmytroHryshyn (approved):

PR #146, opened 2024-03-10
[CLI, Backend] feat: Add codemod list endpoint, tweak list command to work with it, remove sync command, add ts-reset (CDMD-2773)

r4zendev:


apps/cli/src/apis.ts
```diff
@@ -66,3 +67,24 @@ export const getCodemodDownloadURI = async (
 
 	return res.data.link;
 };
+
+export type CodemodListReturn = {
```
DmytroHryshyn, line 71: can we somehow share API types between backend and API consumers?

DmytroHryshyn (approved):

apps/cli/src/apis.ts
r4zendev: pretty sure that if a type includes other types, they all need to be exported. in case with prisms types, we would have to make prisma db connector also a shared package in our monorepo, and then share the types from there. not sure how it affects our bundle sizes and whether this is something that is easy to do. we need to put that into linear and investigate, for now i would not care about it that much. at some point we should bring more modularity and control to our codebase.

PR #147, opened 2024-03-11
[CLI] fix: codemod login experience is broken when user is already logged in - CDMD-2770

hbjORbj:


apps/cli/src/handleLoginCliCommand.ts
```diff
@@ -7,27 +8,54 @@ import { boldText, colorizeText, openURL } from "./utils.js";
 
 const ACCESS_TOKEN_REQUESTED_BY_CLI_KEY = "accessTokenRequestedByCLI";
 
+const routeUserToStudioForLogin = (printer: PrinterBlueprint) => {
+	printer.printConsoleMessage(
+		"info",
+		colorizeText(
+			"Opening the Codemod Studio... Please Sign in there!\n",
```
alexbit-codemod, line 15: Avoid studio. We need to change that soon. 

"Redirecting to Codemod sign-in page..."

@hbjORbj

apps/cli/src/handleLoginCliCommand.ts
```diff
@@ -7,27 +8,54 @@ import { boldText, colorizeText, openURL } from "./utils.js";
...
+			"cyan",
+		),
+	);
+	const success = openURL(
+		`https://codemod.studio/?command=${ACCESS_TOKEN_REQUESTED_BY_CLI_KEY}`,
+	);
+	if (!success) {
+		printer.printOperationMessage({
+			kind: "error",
+			message: "Unexpected error occurred while opening the Codemod Studio.",
```
alexbit-codemod, line 25: Avoid the term studio. 
"Unexpected error occurred while redirecting to sign-in page. Please submit a GitHub issue (sub.co shortlink to github issue template, so we can change it later via dub) or report to codemod.com/community"

alexbit-codemod (commented):
Left some comments. Good to have. Not a blocker.

apps/cli/src/handleLoginCliCommand.ts
```diff
@@ -7,27 +8,54 @@ import { boldText, colorizeText, openURL } from "./utils.js";
...
+		if (username === null) {
+			routeUserToStudioForLogin(printer);
+			return;
+		}
+
 		printer.printConsoleMessage(
 			"info",
 			colorizeText(
-				"Opening the Codemod Studio... Please Sign in with Github!\n",
+				boldText("You are already logged in with the Codemod CLI!"),
```
alexbit-codemod, line 55: "You're already logged in."

apps/cli/src/handleLoginCliCommand.ts
hbjORbj: i think devs will feel more safe knowing that the URL links to github, so I will just leave it as native link

DmytroHryshyn (approved):

PR #150, opened 2024-03-12
[Registry]: Migrate token implementation to use `Prisma` (CDMD-2739)

mirai2k:
- Created Prisma tokens shemas.
- Fully migrated from Sequelize ORM to Prisma ORM.
- Token Service was updated and tested to corresponding Prisma requierments.
- Fixed issue with BigInt.


hbjORbj:
can you see what fails the unit tests?

mohab-sameh:
@all-contributors add @mirai2k for code

apps/backend/prisma/schema.prisma
```diff
@@ -57,3 +57,19 @@ model CodemodVersion {
...
+  encryptedUserId               String @map("euid") @db.VarChar(64)
+  createdAt                     BigInt @map("ca")
+  expiresAt                     BigInt @map("ea")
+  claims                        BigInt @map("c")
+  signature                     String @map("s") @db.VarChar(43)
+}
+
+model TokenRevocation {
+  pepperedAccessTokenHashDigest String @id @map("pathd") @db.VarChar(27)
+  revokedAt                     BigInt @map("r")
```
hbjORbj, line 73:
```suggestion
  revokedAt                     BigInt @map("ra")
```

apps/cli/src/apis.ts
```diff
@@ -15,7 +15,7 @@ export const validateAccessToken = async (
 	accessToken: string,
 ): Promise<Data> => {
 	const response = await Axios.post(
-		"https://backend.codemod.com/validateAccessToken",
+		"http://localhost:8081/validateAccessToken",
```
DmytroHryshyn, line 18: should we move base url to env?

r4zendev: good idea, we don't actually have env currently in CLI, but we can taskify this. i wonder how will it be packaged to the `index.cjs` binary.

apps/backend/prisma/schema.prisma
mirai2k: actually, I keep the same names as we have in [this](https://www.notion.so/codemod/Platform-Authentication-35e8ce29274d48daadce2b890efac1d7) docs. I'll change it soon maybe, we just want right now to make new backend with Prisma works like it was with Sequelize

r4zendev: See references in:
https://github.com/codemod-com/codemod/blob/eac40684c326ab711eb2f7b5e7aadbe78e5cbf47/apps/backend/src/db/buildTokenRevocationsRepository.ts#L64
https://github.com/codemod-com/codemod/blob/eac40684c326ab711eb2f7b5e7aadbe78e5cbf47/apps/backend/src/db/buildTokenMetadataRepository.ts#L82

r4zendev (approved):

DmytroHryshyn (approved):

PR #158, opened 2024-03-13
[codemods] `history.listenBefore()` => `history.block()` - CDMD-2791

hbjORbj:


packages/codemods/history/4/use-block/test/test.ts
```diff
@@ -0,0 +1,41 @@
...
+			console.log('test');
+		});
+		`;
+
+		const output = `
+		import createHistory from 'history/createBrowserHistory';
+
+		const history = createHistory();
+		
+		history.block((location, action) => {
```
DmytroHryshyn, line 24: history block has different args https://github.com/remix-run/history/blob/main/docs/api-reference.md#history.block

packages/codemods/history/4/use-block/src/index.ts
```diff
@@ -0,0 +1,83 @@
...
+				.find(j.Identifier, { name: "listenBefore" })
+				.paths()
+				.at(0);
+
+			if (!identifierPath) {
+				return;
+			}
+
+			// Rename `listenBefore` to `block`
+			identifierPath.replace(j.identifier.from({ name: "block" }));
```
DmytroHryshyn, line 27: can we just mutate the name? 
` identifierPath.node.name = 'block'`

packages/codemods/history/4/use-block/src/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+			const arg = path.value.arguments[0];
+			if (!arg || !j.ArrowFunctionExpression.check(arg)) {
```
DmytroHryshyn, line 30: we dont need `!arg` check here

DmytroHryshyn (dismissed):

PR #160, opened 2024-03-13
[CLI, Backend] feat: Add fuzzy-search command and make it interchangeable by list (CDMD-2768)

r4zendev:
- Adds `codemod search` command that behaves similarly to `codemod list` but allows perfoming fuzzy-find by name on codemod list
- Changes `codemod list` to work exactly the same as search and those two commands to work similarly

### DIFF
--- a/apps/cli/src/executeMainThread.ts
+++ b/apps/cli/src/executeMainThread.ts
@@ -89,8 +89,13 @@ export const executeMainThread = async () => {
 		)
 		.command(
 			"list",
-			"lists all the codemods & recipes in the public registry",
+			"lists all the codemods & recipes in the public registry. can be used similar to search to filter by name",
 			(y) => buildUseJsonOption(buildUseCacheOption(y)),
+		)
+		.command(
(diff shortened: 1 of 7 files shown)


apps/cli/src/executeMainThread.ts
```diff
@@ -192,9 +197,27 @@ export const executeMainThread = async () => {
...
+							"Search term must be at least 2 characters long. Aborting...",
+					});
+					return;
+				}
+				searchTerm = lastArgument;
+			}
+
+			await handleListNamesCommand({
+				printer,
+				name: searchTerm ?? undefined,
```
DmytroHryshyn, line 219: is search case sensitive? should we do searchTerm.toLowerCase() before searching?

DmytroHryshyn: also, lets do searchTerm.trim()

DmytroHryshyn (approved):

DmytroHryshyn (approved):

PR #163, opened 2024-03-13
[codemods] `history.push(path, { state })` => `history.push(path, state)` - CDMD-2792

hbjORbj:


packages/codemods/history/4/replace-state-object/test/test.ts
```diff
@@ -0,0 +1,34 @@
+import assert from "node:assert/strict";
+import { buildApi } from "@codemod-com/utilities";
+import type { FileInfo } from "jscodeshift";
+import { describe, it } from "vitest";
+import transform from "../src/index.js";
+
+describe("history v4 use-back", () => {
```
DmytroHryshyn, line 7: wrong name

packages/codemods/history/4/replace-state-object/test/test.ts
```diff
@@ -0,0 +1,34 @@
...
+	it("should replace history.push(path, { state }) with history.push(path, state)", async () => {
+		const input = `
+		import createHistory from 'history/createBrowserHistory';
+
+		const history = createHistory();
+		
+		history.push('/new-path', { state: { key: 'value' } });
```
DmytroHryshyn, line 14: i've rechecked doc of history v3 and found that this signature is incorrect, it should be like this: 
```
 browserHistory.push({
          pathname: 'str',
          state: { key: value },
        })
      ```
      
      and should be transformed to: 
      
      ```
       browserHistory.push(
          {
            pathname: 'str',
          },
          { key: value }
        )
        ```

Comment (hbjORbj) [reply packages/codemods/history/4/replace-state-object/test/test.ts:14]: Thanks. I also see that the first arg type must be `To`.

`type To = string | Partial<Path>`

and Path looks like

```
export interface Path {
    pathname: Pathname;
    search: Search;
    hash: Hash;
}
```

Ok, so the correct way is the following:
- if the arg is of type string, simply do nothing
- if the arg is of type object, extract `state` and pass its value to 2nd arg and create a new object with any other key/value pairs in the object and pass it to the first arg

DmytroHryshyn: i think we need to do this: 
  - if first arg contains `state` field, extract it to second arg, else - do nothing

DmytroHryshyn (approved):

PR #173, opened 2024-03-14
[CLI, Backend] feat: allow publishing codemods of any engine (CDMD-2826)

r4zendev:
Until now, we only supported publishing of js codemods. We want to allow publishing for all the engines we support, since publishing is now the only way to get codemods into the registry.


mirai2k (approved):
LGTM

PR #194, opened 2024-03-19
[Backend, CLI] feat: tag search, tag table, fix fuzzy search (CDMD-2839, CDMD-2851)

r4zendev:


apps/cli/src/executeMainThread.ts
```diff
@@ -202,21 +202,25 @@ export const executeMainThread = async () => {
...
-						message:
-							"Search term must be at least 2 characters long. Aborting...",
-					});
-					return;
+					throw new Error(
+						"Search term must be at least 2 characters long. Aborting...",
+					);
+				}
+
+				if (argv.tag) {
```
DmytroHryshyn, line 213: i think we should automatically determine if its a tag or name, then user can provide a single search term

DmytroHryshyn (approved):

PR #198, opened 2024-03-20
[Codemods] chore: Update codemodrc files

r4zendev:


DmytroHryshyn (approved):

packages/codemods/react-router/4/hash-router/.codemodrc.json
```diff
@@ -1,15 +1,13 @@
 {
```
DmytroHryshyn, line 1: as i remember, react-router codemods were imported from raja repo https://github.com/rajasegar/react-router-v4-codemods. do we mention the author and include license?

r4zendev: there was no owner field in this config file, so the script i used just blindly updated it to codemod.com. it doesn't have license either. we can change the user to `rajasegar` and add licenses

DmytroHryshyn (approved):

PR #235, opened 2024-03-26
[CLI] feat: ast-grep compatible implementation

r4zendev:
This PR makes it so that ast-grep codemods are executed in the same fashion as our adopted engines. This allows proper messaging like `Processed X files out of Y` and in general makes the execution flow more predictable.


mirai2k (approved):
LGTM

PR #257, opened 2024-03-28
[VSCE] fix: support arguments

DmytroHryshyn:
- restores arguments feature


apps/vsce/src/components/engineService.ts
```diff
@@ -200,16 +204,30 @@ const CODEMOD_ENGINE_NODE_COMMAND = "codemod";
...
 
 	return res.data;
 };
 
+const buildCodemodEntry = (
+	codemod: CodemodListResponse[number],
+): CodemodEntry => {
+	const hashDigest = createHash("ripemd160")
+		.update(codemod.name)
+		.digest("base64url");
```
r4zendev, line 222: we should put that into utilities, i will taskify this. apart from that, LGTM, approved

r4zendev (approved):

PR #263, opened 2024-03-29
[CLI] feat: add check for required args (CDMD-3041)

r4zendev:
- Adjusts logical distribution of sourced and registry codemod runs to "standalone" and "package", because we can infer data from codemod package if it's a valid codemod directory.
- Adds a check for required args and a corresponding message:
image


mirai2k (approved):
lgtm

mirai2k (approved):

PR #280, opened 2024-04-04
[VSCE] fix: multiple codemod run entries are created on one dry-run - CDMD-2696 

hbjORbj:


apps/vsce/src/selectors/selectCodemodRunsTree.ts
```diff
@@ -10,16 +10,10 @@ export const selectCodemodRunsTree = (state: RootState, rootPath: string) => {
...
-				kase.codemodHashDigest !== undefined
-					? state.codemod.entities[kase.codemodHashDigest]?.name ??
-					  kase.codemodName
-					: kase.codemodName;
-
 			return {
 				node: {
 					hashDigest: kase.hash,
-					label,
+					label: kase.codemodName,
```
DmytroHryshyn, line 16: What will happen when we run same codemod multiple times? Will it display all runs, or only 1?

hbjORbj: all runs

hbjORbj: Screenshot 2024-04-04 at 16 59 35
atm, we can distinguish via this

DmytroHryshyn (approved):

PR #288, opened 2024-04-05
chore: add file collection

DmytroHryshyn:
Extracts `inputSnippet` `outputSnippet` states from `snippet` to `file` collection.

### DIFF
--- a/apps/studio/src/pageComponents/main/JSCodeshiftRender.tsx
+++ b/apps/studio/src/pageComponents/main/JSCodeshiftRender.tsx
@@ -10,8 +10,10 @@ import {
 import { useWebWorker } from "~/hooks/useWebWorker";
 import { cn } from "~/lib/utils";
 import { type OffsetRange } from "~/schemata/offsetRangeSchemata";
+import { DEFAULT_TEST_FIXTURE_DIR } from "~/store/getInitialState";
 import { useRangesOnTarget } from "~/store/useRangesOnTarget";
 import { useCodemodOutputStore } from "~/store/zustand/codemodOutput";
+import { useFilesStore } from "~/store/zustand/file";
 import { useLogStore } from "~/store/zustand/log";
 import { useModStore } from "~/store/zustand/mod";
 import { useSnippetStore } from "~/store/zustand/snippets";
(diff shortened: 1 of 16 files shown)


apps/studio/src/pageComponents/main/JSCodeshiftRender.tsx
```diff
@@ -29,7 +31,15 @@ const MonacoDiffEditor = dynamic(
...
 	const { setEvents, events } = useLogStore();
-	const { engine, inputSnippet, afterInputRanges } = useSnippetStore();
+	const { engine, afterInputRanges } = useSnippetStore();
+
+	const { selectFirst } = useFilesStore();
+
+	const inputSnippet =
+		selectFirst({
+			parent: DEFAULT_TEST_FIXTURE_DIR.hashDigest,
+			name: "before.tsx",
```
grzim, line 41: the idea to treat snippets like files is very good.  
how about splitting extension to another property?

apps/studio/src/pageComponents/main/JSCodeshiftRender.tsx
```diff
@@ -29,7 +31,15 @@ const MonacoDiffEditor = dynamic(
...
 export const useCodeDiff = () => {
 	const { setEvents, events } = useLogStore();
-	const { engine, inputSnippet, afterInputRanges } = useSnippetStore();
+	const { engine, afterInputRanges } = useSnippetStore();
+
+	const { selectFirst } = useFilesStore();
+
+	const inputSnippet =
+		selectFirst({
+			parent: DEFAULT_TEST_FIXTURE_DIR.hashDigest,
```
grzim, line 40: maybe we can also add something like 'rootNode' to know exactly to which group a given file belongs?

grzim: or even something like origin which will help us to build stronger type check. In for all files origin snippet names can be only before/after/output what will prevent from typeos

apps/studio/src/store/zustand/file.ts
```diff
@@ -0,0 +1,83 @@
...
+import { INITIAL_STATE } from "../getInitialState";
+
+export type File = {
+	hashDigest: string;
+	name: string;
+	content: string;
+	parent: string | null;
+};
+
+type FileState = {
```
grzim, line 11: maybe FilesStore would be a better name? As this is not a state for a one single file

apps/studio/src/store/zustand/file.ts
```diff
@@ -0,0 +1,83 @@
...
+	files: Record<string, File>;
+	upsertOne: (hashDigest: string, newFile: File) => void;
+	upsertMany: (files: File[]) => void;
+	setAll: (files: File[]) => void;
+	selectAll: (filter?: Partial<File>) => File[];
+	selectById: (hashDigest: string) => File | null;
+	selectFirst: (filter: Partial<File>) => File | null;
+};
+
+function shallowMatch(
```
grzim, line 21: do you have anything against using Ramda for transformations? It is the best functional library in js I know

PR #311, opened 2024-04-09
[CLI, Backend, Utils] feat: keychain implementation, post-install script (CDMD-2464)

r4zendev:
- Adds keychain impl to support writing token to system encrypted storage
- adds post-install script to install esbuild globally for `codemod build` command to work
- fixes esm/cjs issues in utils package
- fixes backend build issue (hopefully)
- adjusts documentation just a bit
- allows changing privacy of the codemod

### DIFF
--- a/apps/docs/codemod-registry/publishing-codemods.mdx
+++ b/apps/docs/codemod-registry/publishing-codemods.mdx
@@ -225,7 +225,7 @@ Once your codemod package is [compatible with Codemod platform](#required-codemo
   codemod login
   ```
   
-  When logging in, you will be redirected to your generated auth token on the web platform. Simply copy your token and enter it into your terminal to login with Codemod CLI.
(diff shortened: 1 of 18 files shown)


mirai2k (approved):

mohab-sameh (approved):

apps/docs/codemod-registry/publishing-codemods.mdx
mohab-sameh, line 1: > We install esbuild globally for you upon installation of Codemod CLI, so you might not need to install it again.

@r4zendev In this case, this whole step is redundant. If user could get codemod login to work, then they have codemod cli installed. This means they have esbuild installed already. It's okay to merge for now, but it's better if we remove the esbuild installation step completely in an upcoming doc improvements PR.

r4zendev: i will test it tomorrow and remove the step from docs if all g.

mirai2k (approved):

PR #315, opened 2024-04-09
[VSCE] fix bucket link for piranha support

hbjORbj:


r4zendev (approved):

r4zendev:
double-checked if rust engine is in the s3 bucket, it is there and updated method's logic should work fine.

PR #328, opened 2024-04-09
[docs] Automate writing Changelog

hbjORbj:


scripts/releaseChangelog.mjs
```diff
@@ -0,0 +1,34 @@
+import fetch from "node-fetch";
```
DmytroHryshyn, line 1: we dont need to import fetch, i think its part of nodejs builtin globals

hbjORbj: addressed

mohab-sameh (approved):

PR #331, opened 2024-04-10
[Filemod]: add ability to move files

DmytroHryshyn:
- adds moveFile API method
- adds checks for removed files/dirs for readDirectory method


r4zendev (approved):
looks good

PR #338, opened 2024-04-10
[Studio] implement codemod run progress bar - CDMD-3092

hbjORbj:
Screenshot 2024-04-10 at 15 05 17


DmytroHryshyn (approved):

DmytroHryshyn:
1. Lets merge it to  main PR `https://github.com/codemod-com/codemod/pull/305` and connect it to codemod run. 
2. User should be able to hide modal (but still see progress), because execution can take a lot of time. Maybe lets show progress on header instead?

grzim:
I already added it in the header so I can plug it there.
@hbjORbj we should call the progress API periodically right?

hbjORbj:
> @grzim we should call the progress API periodically right?

Yes, good point. will do that

PR #372, opened 2024-04-19
Feat/telemetry improvements

DmytroHryshyn:
- adds ability to specify `clientIdentifier` for identifying events that are triggered by different clients - CLI, VSCE, etc; Fixes issue where codemod execution event is triggered twice  - by VSCE and by CLI
 - add util that allows to redact event properties - removing path-like strings
 - add tests for PostHogSender
 - add new event for tracking `codemod/list`  endpoint

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,8 +2,8 @@
 	"editor.defaultFormatter": "biomejs.biome",
 	"editor.formatOnSave": true,
 	"editor.codeActionsOnSave": {
-    "source.fixAll": "explicit",
-    "quickfix.biome": "explicit",
-    "source.organizeImports.biome": "explicit"
-  }
+		"source.fixAll": "explicit",
+		"quickfix.biome": "explicit",
+		"source.organizeImports.biome": "explicit"
+	}
 }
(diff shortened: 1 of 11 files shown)


.vscode/settings.json
```diff
@@ -2,8 +2,8 @@
...
 	"editor.formatOnSave": true,
 	"editor.codeActionsOnSave": {
-    "source.fixAll": "explicit",
-    "quickfix.biome": "explicit",
-    "source.organizeImports.biome": "explicit"
-  }
+		"source.fixAll": "explicit",
+		"quickfix.biome": "explicit",
+		"source.organizeImports.biome": "explicit"
+	}
```
arybitskiy, line 8: We should change tabs to spaces for intendation, imho

apps/backend/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@codemod-com/backend",
-	"version": "0.0.87",
+	"version": "0.0.88",
```
arybitskiy, line 3: is it manual version update?

arybitskiy (approved):

.vscode/settings.json
DmytroHryshyn: i can push this change in next PR

PR #381, opened 2024-04-22
[VSCE] remove codemod package dependency and use globally installed CLI - CDMD-3174, CDMD-3176

hbjORbj:
\

### DIFF
--- a/apps/vsce/src/components/engineService.ts
+++ b/apps/vsce/src/components/engineService.ts
@@ -18,6 +18,7 @@ import {
 	codemodListResponseCodec,
 } from "../codemods/types";
 import type { Configuration } from "../configuration";
+import { CLI_NPM_LINK } from "../constants";
 import type { Container } from "../container";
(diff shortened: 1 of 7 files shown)


apps/vsce/src/components/engineService.ts
```diff
@@ -45,6 +45,9 @@ export const Messages = {
 };
 
 const TERMINATE_IDLE_PROCESS_TIMEOUT = 30 * 1000;
+const currentCLIMajorVersion = 0;
+// TODO: get this info from backend
+const currentCLIMinorVersion = 10;
```
DmytroHryshyn, line 50: i think we can inject it with webpack define plugin instead of hardcoding https://webpack.js.org/plugins/define-plugin/

hbjORbj: i have a better idea. i will ping npm to fetch the latest version.

DmytroHryshyn (approved):

PR #432, opened 2024-04-25
[web app] getURL function no longer needs to be async - CDMD-3205

hbjORbj:


DmytroHryshyn (approved):

apps/frontend/app/(website)/studio/src/hooks/useShareLink.ts
```diff
@@ -55,26 +51,13 @@ export const useShareLink = () => {
...
 
-  const getURL = async (): Promise<URL | null> => {
+  const getURL = (): URL | null => {
     try {
       if (internalContent === null) {
         throw new Error("codemod content not found");
       }
 
       const codemodName = "untitled";
-      //   if (token !== null) {
```
DmytroHryshyn, line 65: why we removed codemod name feature?

hbjORbj: this is for codemod sharing feature in studio, and the name field isn't currently consumed by anywhere by the receiver of the link. So it's overengineering.

PR #438, opened 2024-04-26
[Web app] Dropdown UI in nav bar] - CDMD-3179

hbjORbj:


apps/frontend/components/global/Navigation/DesktopNavigation.tsx
```diff
@@ -112,7 +113,7 @@ export function DesktopNavigationRight(props: {
 }) {
   return (
     <div className="hidden gap-3 lg:flex lg:items-center lg:justify-center">
-      {props.items?.map((item, index) => (
+      {/* {props.items?.map((item, index) => (
```
DmytroHryshyn, line 116: Is this commented intentionally?

DmytroHryshyn (approved):

apps/frontend/components/global/Navigation/DesktopNavigation.tsx
hbjORbj: yes, that was the "Registry" button on the top-right in the nav bar

hbjORbj: but Mohab wants it back actually. i will uncomment it

DmytroHryshyn (approved):

PR #483, opened 2024-05-02
[Backend]: added framework version tag

mirai2k:
- added framework version tag
- bumped version to 0.0.94

### DIFF
--- a/apps/backend/src/services/codemodService.ts
+++ b/apps/backend/src/services/codemodService.ts
@@ -3,7 +3,7 @@ import {
   type CodemodListResponse,
   isNeitherNullNorUndefined,
 } from "@codemod-com/utilities";
-import type { Codemod, Prisma, PrismaClient } from "@prisma/client";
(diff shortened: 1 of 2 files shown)


apps/backend/src/services/codemodService.ts
```diff
@@ -152,6 +153,12 @@ export class CodemodService {
 
     const data: LongCodemodInfo[] = await Promise.all(
       codemods.map(async (codemod) => {
+        let frameworkVersion = null;
+        const match = codemod.name.match(/(\d+(?:\.\d+)*)/);
+        if (match) {
+          frameworkVersion = match[1];
+        }
+
```
arybitskiy, line 161: please move all that code to separate function and use it on both places
also you can simplify with optional chaining
codemod.name.match(/(\d+(?:\.\d+)*)/)?.[0]

arybitskiy (approved):

PR #488, opened 2024-05-02
[studio] fix child missing key error

hbjORbj:


apps/frontend/app/(website)/studio/main/PageBottomPane/Components/side-components.tsx
```diff
@@ -92,7 +92,7 @@ export const AstSection = ({
   engine: KnownEngines;
 }) => {
   return panels.filter(isVisible).map((panel, i, { length }) => (
-    <>
+    <React.Fragment key={panel.relatedAST}>
```
DmytroHryshyn, line 95: is `panel.relatedAST` a string?

DmytroHryshyn (approved):

apps/frontend/app/(website)/studio/main/PageBottomPane/Components/side-components.tsx
hbjORbj: it's a number, so it should be fine

PR #514, opened 2024-05-08
[CLI, utilities] feat: init during publish (CDMD-3283)

r4zendev:
Upon merging this PR, #506 can be closed. (@DmytroHryshyn)

- `codemod init` questions are adjusted if main file/folder path is provided
- When user tries to run `codemod publish` on a directory, we will ask him to provide a path to the main file in that directory and use this directory as a default path to run `codemod init` on.
- When user tries to run it on a concrete file, we do the same, but get the default directory based on where this file is located.


mirai2k (approved):
LGTM

PR #542, opened 2024-05-10
[CLI] fix, chore: prepublish script, bump version

r4zendev:


apps/cli/package.json
```diff
@@ -65,7 +65,7 @@
     "package": "pkg --compress GZip .",
     "test": "TEST=1 vitest run",
     "coverage": "TEST=1 vitest run --coverage",
-    "prepublishOnly": "pnpm run build",
+    "prepublishOnly": "turbo run build --filter codemod",
```
DmytroHryshyn, line 68: will it automatically build dependent packages?

r4zendev: yes, this is the purpose of this change.
previously, it would not, and it caused stale code to appear in npm.
also, i have built the cli previously using `pnpm --filter codemod build`, we should do `pnpm build --filter codemod` now in order to achieve the same behaviour of building dependencies.

DmytroHryshyn (approved):

DmytroHryshyn (approved):

PR #625, opened 2024-05-17
Initial workflow PR

arybitskiy:
**Please check if the PR fulfills these requirements**:

-   [x] The commit message follows [our code of conduct](https://github.com/codemod-com/codemod-registry/blob/main/CODE_OF_CONDUCT.md) and [contributing guidelines](https://github.com/codemod-com/codemod-registry/blob/main/CONTRIBUTING.md)
-   [ ] Tests for the changes have been added (for bug fixes/features)
-   [ ] Docs have been added / updated (for bug fixes / features)

This is potentially new workflow engine
Example of usage is in `packages/workflow/sample.workflow.ts`
There is documentation with JSDoc in `packages/workflow/src/compose.ts`
General design:
Every call to a function is lazy and returns Promise which is not executed until you call `await promise`.


apps/cli/src/commands/run.ts
```diff
@@ -99,6 +100,16 @@ export const handleRunCliCommand = async (
...
+    try {
+      if (await isWorkflowFile(nameOrPath)) {
+        await runWorkflowFile(nameOrPath);
+        return;
+      }
+    } catch (e: any) {
+      console.error(e);
+      console.log(e?.stack);
+      return;
+    }
```
r4zendev, line 112: since you have added boilerplate to run codemods that have `engine: workflow`, can this be removed?

packages/workflow/sample.ts
r4zendev, line 1: should be removed from the PR?

packages/workflow/sample.workflow.ts
r4zendev, line 1: should be removed from the PR?

packages/workflow/samples/react-styles-to-css-modules/ButtonAfter.tsx
r4zendev, line 1: let's rename folder to `examples`, more conventional

packages/workflow/src/PLazy.ts
```diff
@@ -0,0 +1,66 @@
...
+    onFulfilled:
+      | ((resolvedValue: ValueType) => TResult1 | PromiseLike<TResult1>)
+      | null
+      | undefined,
+    onRejected:
+      | ((error: any) => TResult2 | PromiseLike<TResult2>)
+      | null
+      | undefined,
+  ) {
+    // TODO: Use `??=` when targeting Node.js 16.
```
r4zendev, line 56: we are targeting node >16, so either remove the comment or use nullish assignment operator

packages/workflow/src/astGrep/astGrep.ts
```diff
@@ -0,0 +1,114 @@
...
+const astGrepHelpers = {
+  replace,
+  map,
+};
+
+type AstGrepHelpers = typeof astGrepHelpers;
+
+/**
+ *
+ * @param query Some description here
```
r4zendev, line 23: add description?

packages/workflow/src/astGrep/astGrep.ts
```diff
@@ -0,0 +1,114 @@
...
+ * @example
+ * ### Pass parameter as string
+ * ```ts
+ *   await astGrep("import React from 'react'");
+ * ```
+ * @example
+ * ```ts
+ *   await astGrep`import React from 'react'`;
+ * ```
+ * @example asdasd
```
r4zendev, line 33: asdasd

packages/workflow/src/astGrep/astGrep.ts
```diff
@@ -0,0 +1,114 @@
...
+  callback?: (helpers: AstGrepHelpers) => Promise<void> | void,
+) {
+  const innerParentContext = getParentContext();
+
+  const grep =
+    typeof query === "object"
+      ? Array.isArray(query)
+        ? (query.join("") as string)
+        : (query as NapiConfig)
+      : query;
```
r4zendev: if possible, try to avoid nested ternary and use imperative style, as it's much more readable

packages/workflow/src/git/branch.ts
```diff
@@ -0,0 +1,73 @@
...
+      const voidOrPromise = callback(wrapHelpers(helpers, context));
+      if (voidOrPromise instanceof Promise) {
+        voidOrPromise
+          .then(() => resolve(wrapHelpers(helpers, context)))
+          .catch(reject);
+      }
+    } else {
+      context().then(resolve).catch(reject);
+    }
+  }) as PLazy<Helpers> & Helpers;
```
r4zendev, line 67: later, as an improvement, this can be abstracted out IMO. i see this pattern used across lots of files

packages/workflow/src/git/git.ts
r4zendev, line 1: generally, for git operations, it could be fine to just use a library, like `simple-git`. but if you prefer it that way, we can keep it like that for now. later by using a library and wrapping its methods under your PLazy approach, we can provide literally native git experience out of the box and allow users to use it inside of their workflows. specially if we can tree-shake our library of choice.

packages/workflow/src/fs.ts
```diff
@@ -0,0 +1,178 @@
...
+          );
+        } else {
+          await fs.writeFile(filepath, JSON.stringify(updater, null, 2));
+        }
+      }
+    },
+  });
+};
+
+// export const jsFiles = async <T>(
```
r4zendev, line 141: comments in this file. are they needed?

packages/workflow/src/isWorkflowFile.ts
r4zendev, line 1: we should get rid of this at some point as well. can we detect if it's a workflow file based on whether the workflow library was used? are we required to bundle (transpile) the workflow before running it via CLI?

packages/workflow/src/jsFiles.ts
```diff
@@ -0,0 +1,98 @@
...
+      const voidOrPromise = context(callback);
+      if (voidOrPromise instanceof Promise) {
+        voidOrPromise
+          .then(() => resolve(wrapHelpers(helpers, context)))
+          .catch(reject);
+      }
+    } else {
+      context().then(resolve).catch(reject);
+    }
+  }) as PLazy<Helpers> & Helpers;
```
r4zendev, line 90: again pattern

packages/runner/src/runCodemod.ts
```diff
@@ -186,6 +187,17 @@ export const runCodemod = async (
...
     });
 
+  if (codemod.engine === "workflow") {
+    const codemodSource = await readFile(codemod.indexPath, {
+      encoding: "utf8",
+    });
+    const transpiledSource = codemod.indexPath.endsWith(".ts")
+      ? transpile(codemodSource.toString())
+      : codemodSource.toString();
+    runWorkflowCodemod(transpiledSource, safeArgumentRecord, console.log);
```
r4zendev, line 197: await?

packages/workflow/src/contexts.ts
```diff
@@ -0,0 +1,105 @@
...
+  invariant(ctx, "No ast-grep node context found");
+  return ctx;
+};
+
+export const getCwdContext = () => {
+  const cwd = cwdContext.getStore();
+  if (!cwd) {
+    return { cwd: process.cwd() };
+  }
+  return cwd;
```
r4zendev, line 80: `return cwd ?? { cwd: cwd() }`

r4zendev (approved):
looks fine by me, vercel deployment fails for some reason, maybe should rebase and check if it succeeds just in case

PR #713, opened 2024-05-27
[CLI] feat: smarter include/exclude options (CDMD-3450, CDMD-3451, CDMD-3462, CDMD-3470, CDMD-3448)

r4zendev:
- Now showing the target that codemod is going to run on in the run summary
- Codemod include and exclude patterns became smarter. If they are provided by user through options or through codemod config options, include patterns take priority. If they are determined by engine or set by ast-grep configuration, exclude patterns provided by user through arguments will take over. Depending on which one has the priority, the other list will be filtered to not contain patterns from the prioritized list.
- Added `dist`, `build`, `.next` folders to default excluded patterns
- Removed .d.ts files from ignored by default
- Now we assume that path is a directory and should be formatted as `**/path/**/*.*` glob pattern if user supplies it with a trailing slash
- Now supporting specifying target path through positional argument without needing to prepend `-t`
- Now populating exclude patterns from root `.gitignore` file in the target directory
- Improved argv options in general, now the help menu looks like this:
image
which is much more comprehensive for the user.

Some examples:
image

### DIFF
--- a/packages/runner/src/schemata/flowSettingsSchema.ts
+++ b/packages/runner/src/schemata/flowSettingsSchema.ts
@@ -14,6 +14,7 @@ import {
 
 export const DEFAULT_EXCLUDE_PATTERNS = [
   "**/node_modules/**/*.*",
+  "**/*.d.ts",
   "**/.next/**/*.*",
   "**/dist/**/*.*",
   "**/build/**/*.*",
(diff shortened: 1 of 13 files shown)


packages/runner/src/schemata/flowSettingsSchema.ts
```diff
@@ -11,19 +12,26 @@ import {
   string,
 } from "valibot";
 
-export const DEFAULT_EXCLUDE_PATTERNS = ["**/node_modules/**/*.*", "**/*.d.ts"];
```
DmytroHryshyn, line 14: why we removed *d.ts from default exclude? is there use case where we want to run codemod on d.ts files?

DmytroHryshyn (approved):

PR #743, opened 2024-05-28
Mok/recipe run summary and patterns

r4zendev:
- Fix recipe run summary to include summaries for all the sub-codemods
- Replace `**/*` pattern with `**/*.*` so that it doesn't override ignore patterns
- Added type-checking step for CLI building and fixed remaining type errors
- Fixed a critical bug where codemods would not process all the available paths because worker thread manager would exit early. Now we are waiting until path generator drains.
- Improved perf by moving summary printing into `runCodemod` function, which allowed removing redundant call to `buildPatterns`.
- Fixes broken recipe codemod execution progress due to console messages being suppressed for some reason


packages/runner/src/runCodemod.ts
```diff
@@ -558,6 +635,10 @@ export const runCodemod = async (
         }
 
         if (message.kind === "finish") {
+          // onPrinterMessage({
```
DmytroHryshyn, line 638: do we need this commented block of code?

DmytroHryshyn (approved):

PR #844, opened 2024-06-10
[docs] add codemod publish source option

mohab-sameh:
Adds `codemod publish` `--source` option.

### DIFF
--- a/apps/docs/deploying-codemods/cli.mdx
+++ b/apps/docs/deploying-codemods/cli.mdx
@@ -78,10 +78,10 @@ To do so, use the `publish` command inside the codemod package directory:
 codemod publish
 ```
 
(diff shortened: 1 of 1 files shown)


apps/docs/deploying-codemods/cli.mdx
```diff
@@ -72,10 +72,18 @@ Can be used to publish a codemod to Codemod Registry.
...
 ```bash
 codemod publish
 ```
 
+Alternatively, you can publish a codemod from a local <Tooltip tip="This file can be a simple transform file using any of the supported codemod engines.">transform file</Tooltip>:
+
+```bash
+codemod publish [path-to-file] --source
+```
+
```
r4zendev, line 86: it can be not only a transform file, but also a codemod-compatible package that resides somewhere outside of user current directory. this is slightly misleading.

alexbit-codemod (approved):

PR #1014, opened 2024-07-01
feat: add codemod-utils

DmytroHryshyn:
Add codemod-utils package, that includes utils for multiple kinds of nodes and for react framework: 
   - callExpression
   - function
   - class
   - import 
   - react 
   - global utils


packages/codemod-utils/src/class.ts
```diff
@@ -0,0 +1,37 @@
+import type {
+  ASTPath,
+  ClassDeclaration,
+  ClassMethod,
+  ClassProperty,
+  JSCodeshift,
+} from "jscodeshift";
+
+export const getClassMethod = (
```
DmytroHryshyn, line 9: Should we keep class utils? It looks kinda useless to me

packages/codemod-utils/src/function.ts
```diff
@@ -0,0 +1,63 @@
...
+  ArrowFunctionExpression,
+  Collection,
+  FunctionDeclaration,
+  FunctionExpression,
+  JSCodeshift,
+} from "jscodeshift";
+
+import { getDefaultExport, isDefinedInGlobalScope } from "./global.js";
+
+export type FunctionLike =
```
DmytroHryshyn, line 12: Im still not sure if its a good idea to combine multiple node types into single type. From usability perspective it looks good, because we can have same utils for different kinds of functions (Declarations and ArrowFunctions).

packages/codemod-utils/src/importDeclaration.ts
```diff
@@ -0,0 +1,313 @@
...
+ * @param j - The JSCodeshift API.
+ * @param importDeclaration - The import declaration to search.
+ * @param name - The imported name of the named import to retrieve.
+ * @returns The import specifier if found, otherwise null.
+ */
+export const getNamedImport = (
+  j: JSCodeshift,
+  importDeclaration: ASTPath<ImportDeclaration>,
+  name: string,
+): ImportSpecifier | null =>
```
DmytroHryshyn, line 22: Should we return ASTPath or the node itself here and in other utils?

grzim (approved):

PR #1026, opened 2024-07-03
add Codemod2.0 guide doc

mohab-sameh:


apps/docs/guides/building-codemods/codemod2.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+export async function workflow({ files }: Api) {
+  await files()
+	  **.jsFam()
+}**
+```
+
+### Step 3: Use ast-grep for pattern detection.
+
+Let's use [ast-grep](https://ast-grep.github.io/) which is built-in into our workflow engine
```
alexbit-codemod, line 53: from here, we should link to workflow engine doc, astGrep section, that explains different ways to use astGrep API in workflow engine... there are different ways we can use this API depending on what the original ast grep codemod looks like... this improvement can happen later once the doc for workflow engine is ready... @arybitskiy , @mohab-sameh

mohab-sameh: Sounds good

apps/docs/guides/building-codemods/codemod2.mdx
```diff
@@ -0,0 +1,177 @@
...
+	}
+```
+
+### Step 5: Run and test the codemod locally.
+
+Let's check it out:
+
+```bash
+npx codemod@latest mui-named-import-to-direct-import/src/index.ts \
+--OPENAI_API_KEY=$OPEN_API_KEY \
```
alexbit-codemod, line 98: is this --OPENAI_API_KEY hardcoded? 
or should we provide instructions to add that in codemodrc.json?
@arybitskiy

alexbit-codemod (approved):

PR #1079, opened 2024-07-15
[Frontend, backend] feat: temporary fix for building misformed code in studio, spinning indicators, add hidden field, add hash to codemod name if published from studio (CDMD-3789, CDMD-3790, partly CDMD-3791, CDMD-3784)

r4zendev:
- fix spinning indicators
- temporary fix for building code
- add hidden field
- add random hash to codemod name
- change command to say user needs to login
- change command to use `@latest` when `npm` tab is open, otherwise use `pnpx`
- adds a way to display hidden codemods and user's own codemods
- additional refactoring (`getUserDataOrLogin`, `transpileTs`)
- fixes archive with codemod downloaded from studio not having all test fixtures

image


apps/backend/src/publishHandler.ts
```diff
@@ -278,6 +284,7 @@ export const publishHandler: RouteHandler<{
           engine: codemodRc.engine,
           applicability: codemodRc.applicability,
           verified: isVerified,
+          hidden: isPublishedFromStudio,
```
DmytroHryshyn, line 287: What is difference between `hidden` and `private`?

r4zendev: hidden is meant to hide the codemod from being visible in the registry/CLI by default. the codemods that are published from studio by using `Run via CLI` feature are considered as temporary versions of the codemods that are not meant to be seen implicitly. so we hide them by default and give user an option to view them (registry switch is in progress)

r4zendev: even if the codemod is private, it will show up in the registry/CLI for the user.

DmytroHryshyn (approved):

PR #1120, opened 2024-07-22
feat(backend, cli, utilities, printer, runner): refactoring, upload entire source code of the codemod, small fixes (CDMD-3525)

r4zendev:
- unused code deprecation
- refactoring (imports, file names, folders)
- rewrite of key experiences in CLI
- upload entire codemod source to the S3
- tests fixes
- cover crucial parts with valibot schema validations
- remove build step for codemods from the user side
- unify tsconfig experiences and add subpath imports to the codebase (#import-path.js)
- adds dry-run feature for filemod codemods


apps/cli/src/main.ts
```diff
@@ -273,27 +249,27 @@ export const executeMainThread = async () => {
...
-
-        return executeCliCommand(() =>
-          handleBuildCliCommand({
-            printer,
-            source: args.source ?? process.cwd(),
-          }),
-        );
-      },
-    )
+    // .command(
```
DmytroHryshyn, line 252: Why was it commented?

r4zendev: to be removed, this command is no longer needed.

packages/deprecated/src/vsce/piranha-subscription.ts
```diff
@@ -0,0 +1,55 @@
...
+//   vscode.commands.registerCommand(
+//     "codemod.executeAsPiranhaRule",
+//     async (configurationUri: vscode.Uri) => {
+//       const fileStat = await vscode.workspace.fs.stat(configurationUri);
+//       const configurationUriIsDirectory = Boolean(
+//         fileStat.type & vscode.FileType.Directory,
+//       );
+
+//       if (!configurationUriIsDirectory) {
+//         throw new Error(
```
arybitskiy, line 11: do you want to remove those?

arybitskiy (approved):

packages/deprecated/src/vsce/piranha-subscription.ts
r4zendev: yeah, we don't support piranha anymore @arybitskiy

r4zendev: at least basically no effort has been put into this during the last 6 months or so

r4zendev: i am moving the deprecated code to `deprecated` package that is not in our turbo pipeline, hence is not being built and does not affect deployments in any way. i expect it to stay there for some time until we reach stability so that we don't lose it in case we suddenly want to support those features again

apps/backend/src/publishHandler.ts
```diff
@@ -1,38 +1,53 @@
...
+import { environment } from "./util.js";
 
 export type PublishHandlerResponse = ApiResponse<PublishResponse>;
 
 export const publishHandler: RouteHandler<{
   Reply: PublishHandlerResponse;
 }> = async (request: UserDataPopulatedRequest, reply) => {
+  const unpackPath = join(
+    homedir(),
+    ".codemod",
```
alexbit-codemod, line 46: @mohebifar

PR #1138, opened 2024-07-24
feat(studio): remove legacy ai button

grzim:
image


alexbit-codemod:
nits: since the clean up also impacts the UI for users (simplifies the UI), it will be considered as feat: per conventional commits. if it was not in the UI and you just wanted to clean up the code, it would be chore. if fixed this myself.

PR #1175, opened 2024-08-01
fix(cli,backend): patch a bug that incorrectly extracts codemod name and version

r4zendev:
running:
image

produces:
image

### DIFF
--- a/packages/utilities/src/functions/formatting.ts
+++ b/packages/utilities/src/functions/formatting.ts
@@ -101,3 +101,21 @@ export const buildCodemodSlug = (name: string) =>
     .replaceAll("@", "")
     .split(/[\/ ,.-]/)
     .join("-");
+
+export const extractNameAndVersion = (val: string) => {
+  const lastAtSymbolIndex = val.lastIndexOf("@");
+  const sliceUpToLastAtSymbol = val.substring(0, lastAtSymbolIndex);
+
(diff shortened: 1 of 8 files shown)


packages/utilities/src/functions/formatting.ts
```diff
@@ -101,3 +101,21 @@ export const buildCodemodSlug = (name: string) =>
...
+    return {
+      name: val,
+      version: null,
+    };
+  }
+
+  return {
+    name: sliceUpToLastAtSymbol,
+    version: val.substring(lastAtSymbolIndex + 1),
+  };
```
mohebifar, line 120: how about using a regex here?

```ts
val.match(/^(?<name>@?[^@]+)(?:@(?<version>[0-9]+\.[0-9]+\.[0-9]+))?$/).groups
```

r4zendev: i'd argue that overusing regex where it's unnecessary is damaging readability of the code and also might affect performance in more complex examples. i prefer to stay away from using regular expressions as much as possible.

r4zendev: also version is not necessarily 3 digits separates by dots. we also support string versions such as "latest", "alpha" etc.

mirai2k (approved):

PR #1207, opened 2024-08-09
fix(cli): allow targeting single file, preserve source scope

r4zendev:
- Sets target as `dirname(target)` and does `include = [...include, basename(target)` when target is a file.
- Passes entire codemod source through worker thread message to avoid constructing new annonymous function that loses the scope. Previously I've done that in attempt to avoid calling`getTransformer` on every processed file, however I overlooked that this optimization would cause this bug and turn out to be inefficient, since calling `new Function()` is probably way heavier than just extracting the transformer. The architecture can be improved here in some way, for example passing the transformer *name* with the message, but for now let's leave it bare min so that it works as before.

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -4,7 +4,7 @@
   "imports": {
     "#*": "./src/*"
   },
-  "version": "0.13.2",
+  "version": "0.13.3",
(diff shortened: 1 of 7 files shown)


mirai2k (approved):
lgtm

PR #1219, opened 2024-08-13
fix(codemod): ignore type imports in replace-act-import

henryqdineen:
#### 📚 Description
We ran into an issue with the `replace-act-import` codemod where the `act` import gets added to a type import like

```diff
- import { act } from 'react-dom/test-utils';
- import type { ReactNode } from 'react';
+ import type { ReactNode, act } from 'react';
```

My fix here is to filter out any `importKind` that is not `value` or `undefined`

#### 🧪 Test Plan
I added a unit test in the PR

N/A

### DIFF
--- a/packages/codemods/react/19/replace-act-import/src/index.ts
+++ b/packages/codemods/react/19/replace-act-import/src/index.ts
@@ -101,6 +101,9 @@ export default function transform(
 
       const existingReactImportCollection = root.find(j.ImportDeclaration, {
         source: { value: "react" },
+        importKind(value) {
+          return value === "value" || value == null;
+        },
         specifiers: [{ type: "ImportSpecifier" }],
       });
 
--- a/packages/codemods/react/19/replace-act-import/__testfixtures__/existing-type-import.input.js
+++ b/packages/codemods/react/19/replace-act-import/__testfixtures__/existing-type-import.input.js
@@ -1,0 +1,4 @@
+import type { FC } from "react";
+import { act } from "react-dom/test-utils";
+
+act();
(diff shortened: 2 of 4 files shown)


alexbit-codemod:
Thanks for your contribution! The team will review and merge it soon.

packages/codemods/react/19/replace-act-import/src/index.ts
```diff
@@ -101,6 +101,9 @@ export default function transform(
 
       const existingReactImportCollection = root.find(j.ImportDeclaration, {
         source: { value: "react" },
+        importKind(value) {
+          return value === "value" || value == null;
```
r4zendev, line 105: eqeqeq (`===`), other than that LGTM, thanks
can you bump the patch version in `.codemodrc.json`, please?

r4zendev (approved):

mohab-sameh:
@all-contributors add @henryqdineen for code and bug fix.

mohab-sameh:
@henryqdineen I updated your PR description with a line for accepting the CLA. Please check and accept the CLA before merging. Sorry for the hassle. We configured a CLA bot for the future so that this process is automated.

packages/codemods/react/19/replace-act-import/src/index.ts
henryqdineen: The type of `importKind` is `"type" | "value" | "typeof" | undefined` so the `==` was to also match undefined. 

I'm honestly not sure if realistically any parser used with jscodeshift would create an AST with an empty `importKind`

PR #1223, opened 2024-08-14
feat(cli): add no-interactive flag

arybitskiy:
#### 📚 Description
Added no-interactive flag for running cli (for example do not ask which recipes to run)

### DIFF
--- a/apps/cli/src/flags.ts
+++ b/apps/cli/src/flags.ts
@@ -142,5 +142,10 @@ export const buildRunOptions = <T>(y: Argv<T>) => {
       type: "boolean",
       description: "Run codemod in the cloud",
       default: false,
+    })
+    .option("no-interactive", {
+      type: "boolean",
+      description: "Disable interactive mode",
+      default: false,
     });
 };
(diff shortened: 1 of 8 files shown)


DmytroHryshyn (approved):

apps/cli/src/flags.ts
```diff
@@ -142,5 +142,10 @@ export const buildRunOptions = <T>(y: Argv<T>) => {
       type: "boolean",
       description: "Run codemod in the cloud",
       default: false,
+    })
+    .option("no-interactive", {
+      type: "boolean",
+      description: "Disable interactive mode",
+      default: false,
     });
```
r4zendev, line 150: we can use already existing "mode" flag for that

arybitskiy: it is a little different, `mode` is about output
here i'm disabling input, for example when you run `recipe codemod` programmatically and user can't make those choices

r4zendev: is there a case where we would want to use both "json" mode and "non-interactive"?

r4zendev (approved):

PR #1242, opened 2024-08-16
fix(codemod): support aliased act() import

henryqdineen:
#### 📚 Description
This fixes an edge cast that we found in our codebase where the `act` named import was being aliased.

```diff
-import { ComponentType, ReactNode } from 'react';
-import { act as domAct } from 'react-dom/test-utils';
+import { ComponentType, ReactNode, act } from 'react';
```

The fix is to copy over all `act` specifiers instead of assuming that the new specifier is not aliased. I understand that it's impossible to handle all edge cases but this one _seems_ straightforward. As an alternative you could check that `local.name == 'act'` and skip transforming the import. That would prevent introducing broken code but limit the capabilities of the codemod.

#### 🧪 Test Plan
Added a unit test. 

N/A

### DIFF
--- a/packages/codemods/react/19/replace-act-import/__testfixtures__/aliased-import.input.js
+++ b/packages/codemods/react/19/replace-act-import/__testfixtures__/aliased-import.input.js
@@ -1,0 +1,3 @@
+import { act as reactAct } from "react-dom/test-utils";
+
+reactAct();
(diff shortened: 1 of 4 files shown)


alexbit-codemod:
thank you so much for your contribution @henryqdineen. we will review and merge this soon.

mohebifar (approved):
Looks good! Thanks!

alexbit-codemod:
this is old! 
the issue had been fixed in our new react codemods here https://github.com/codemod/react-codemod/tree/main/codemods/replace-act-import

thanks again for reporting this issue, moving forward, any codemod-specific issues should be reported in the repo that hosts the codemod source. this repo is for Codemod cli. thanks

PR #1281, opened 2024-08-26
fix(cli): bugfix for stdout being blocked

r4zendev:


mohebifar (commented):
if you want to release a new version, you should run `pnpm changeset` in the root and check in the generated `.md` file.

mohebifar (approved):

PR #1286, opened 2024-08-27
feat: pkg.pr.new

Aslemammad:
This PR adds continuous releases to codemod! 

Before running the workflow, we'd have to install the [app](https://github.com/apps/pkg-pr-new)!

### DIFF
--- a/.github/workflows/cr.yml
+++ b/.github/workflows/cr.yml
@@ -1,0 +1,30 @@
+name: Continuous Releases
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


r4zendev:
Hey, @Aslemammad, thanks for that, however we already use a tool that does the releases for us. We use [changesets](https://github.com/changesets/changesets). Not sure what is this for. Can you explain how it brings any benefits over what we already use during dev process?

r4zendev (approved):

mohab-sameh:
@all-contributors add @Aslemammad for infra

Aslemammad:
Thank you so much @mohab-sameh ❤️

PR #1331, opened 2024-09-24
feat(cli,runner): support for exported parser and include cjs/esm-specific extensions

r4zendev:
- Adds support for `export const parser =` and `module.exports.parser =`
- Adds `.mts`, `.mjs`, `.cts`, `.cjs` extensions to the default included patterns when running a javascript-based codemod.

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -4,7 +4,7 @@
   "imports": {
     "#*": "./src/*"
   },
-  "version": "0.14.0",
+  "version": "0.14.1",
   "description": "A codemod engine for Node.js libraries (jscodeshift, ts-morph, etc.)",
   "type": "module",
   "exports": null,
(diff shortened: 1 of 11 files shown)


apps/cli/package.json
```diff
@@ -4,7 +4,7 @@
   "imports": {
     "#*": "./src/*"
   },
-  "version": "0.14.0",
+  "version": "0.14.1",
```
mohebifar, line 7: Let's use changset for version bumps

packages/runner/src/source-code.ts
```diff
@@ -24,9 +26,69 @@ export const isESMExtension = (path: string) =>
...
+        temporaryLoadedModules.set(hashDigest, {
+          transformer: alreadyLoaded?.transformer ?? null,
+          parser,
+        });
+        return parser;
+      } catch (err) {
+        return null;
+      }
+    }
+  };
```
mohebifar, line 74: Can we somehow avoid performing this twice? We're already doing the bulk of this in `getTransformer`?

mohebifar (approved):

PR #1414, opened 2024-12-17
fix: typescript/remove-public-modifier: keep private/protected

pjonsson:
#### 📚 Description

Only remove public modifier, keep
any private/protected modifiers.

#### 🔗 Linked Issue

No issue.

#### 🧪 Test Plan

`pnpm test`

Bug fix, no functional change compared to the documentation.


pjonsson:
@mohab-sameh ok, I will make a PR there instead.

I can't run https://codemod.com/registry/classnames-to-clsx (Request fails with 403 when fetching classnames-to-clsx), where should that kind of stuff be reported?

mohab-sameh:
@pjonsson please report the issue in the commons repo as well. We'll update our issue reporting guidelines soon to clarify what gets reported where.

PR #1444, opened 2025-01-10
fix: hide example heading in markdown purview in registry card

amirabbas-gh:
#### 📚 Description

This PR improves the UI/UX of the registry page by hiding the "example" heading in the `getDescriptionShortText` function. Removing this unnecessary heading results in a cleaner and more streamlined user experience.


mohab-sameh:
somewhat related:
could we also bypass code blocks:
image
or rather bypass anything that is not normal text?

mohab-sameh (approved):

PR #1474, opened 2025-02-04
feat(packages/codemod-utils): add tsup and CJS for require support in codemod-utils

amirabbas-gh:
#### 📚 Description

This PR adds `tsup` as a build tool and includes CommonJS (`.cjs`) support in **codemod-utils**.  
With this change, the package can now be used with both `import` and `require`, improving compatibility.  

#### 🧪 Test Plan

- Built the package using `tsup` and verified that both ESM (`.js`) and CommonJS (`.cjs`) outputs were generated correctly.  
- Tested importing the package in both `import` and `require` environments to ensure compatibility.

### DIFF
--- a/packages/codemod-utils/tsup.config.ts
+++ b/packages/codemod-utils/tsup.config.ts
@@ -1,0 +1,13 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
+  entry: ["src/index.ts"],
+  format: ["esm", "cjs"],
+  dts: {
+    compilerOptions: {
+      incremental: false, // Fix for incremental error
+    },
+  },
+  clean: true,
+  sourcemap: true,
+});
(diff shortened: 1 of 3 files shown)


packages/codemod-utils/tsup.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
+  entry: ["src/index.ts"],
+  format: ["esm", "cjs"],
+  dts: {
+    compilerOptions: {
+      incremental: false, // Fix for incremental error
```
mohebifar, line 8: Why is this needed?

amirabbas-gh: @mohebifar From my searches, I understand that we need a `tsup.config` for this package, and "incremental" on line 8 is for handling some errors.

mohebifar (approved):

PR #1549, opened 2025-06-19
feat(docs): add new cli docs

mohab-sameh:
New Codemod CLI docs 🥳 

Merge when new Codemod CLI (~Butterflow~) is released. @mohebifar If you prefer the name Codemod Workflows instead, we can use that. I prefer that to avoid confusion with the legacy CLI + the scope of this tool is bigger than the context of a CLI.

FYI, reviewing would be really helpful since some of the tooling is still TBD so getting we might want to change things based on where we see the product once the CLI is released.

https://github.com/user-attachments/assets/90068536-55d5-40ed-86f5-fc0fd1d0b2bb

### DIFF
--- a/apps/docs/workflows.mdx
+++ b/apps/docs/workflows.mdx
@@ -1,0 +1,499 @@
+---
+title: 'Codemod CLI'
+description: Run, validate, and automate code-modification workflows locally or in the cloud.
+icon: 'code'
+---
+
+---
+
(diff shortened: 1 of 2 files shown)


apps/docs/workflows.mdx
```diff
@@ -0,0 +1,499 @@
+---
+title: 'Codemod CLI'
```
mohebifar: I think the title of the page should be Codemod Workflow. We need a master doc about Codemod CLI and Codemod workflow is one of the sub-pages.

Another feature of the Codemod CLI that we need to cover is [jssg](https://github.com/codemod-com/codemod/tree/main/crates/cli#jssg-javascript-ast-grep)

alexbit-codemod (approved):

PR #1565, opened 2025-06-27
refactor: return 410 GONE for deprecated endpoints and restrict old CLI to only `codemod run`

amirabbas-gh:
#### 📚 Description

This PR introduces a breaking change by:

- Refactoring deprecated backend endpoints to return a `410 GONE` HTTP error.
- Disabling all old CLI commands except for `codemod run`.

These changes aim to clean up legacy functionality, reduce confusion for users, and encourage the use of supported interfaces only.

#### 🧪 Test Plan

- Verified that requests to deprecated endpoints return a `410 GONE` response with the correct message.
- Ran old CLI commands to confirm that only `codemod run` is allowed.

### DIFF
--- a/.changeset/three-bats-glow.md
+++ b/.changeset/three-bats-glow.md
@@ -1,0 +1,7 @@
+---
+"@codemod-com/backend": major
+"codemod": major
+"@codemod-com/api-types": patch
+---
+
+Refactored deprecated endpoints to return a 410 GONE error, disabled all CLI commands except codemod run
--- a/.changeset/three-bats-glow.md
+++ b/.changeset/three-bats-glow.md
@@ -1,6 +1,5 @@
 ---
-"@codemod-com/backend": major
-"codemod": major
+"codemod": patch
 "@codemod-com/api-types": patch
 ---
 
--- a/apps/backend/src/publishHandler.ts
+++ b/apps/backend/src/publishHandler.ts
@@ -241,7 +241,8 @@ export const publishHandler: RouteHandler<{
           "This endpoint is no longer supported. Please use the new CLI instead.",
         error: GONE,
       });
-    } else if (!semver.gt(version, latestVersion.version)) {
+    }
+    if (!semver.gt(version, latestVersion.version)) {
       return reply.code(400).send({
         error: CODEMOD_VERSION_EXISTS,
         errorText: `Codemod ${name} version ${version} is lower than the latest published or the same as the latest published version: ${latestVersion.version}`,
--- a/apps/backend/src/publishHandler.ts
+++ b/apps/backend/src/publishHandler.ts
@@ -12,6 +12,7 @@ import {
   CODEMOD_CONFIG_INVALID,
   CODEMOD_NAME_TAKEN,
   CODEMOD_VERSION_EXISTS,
+  GONE,
   INTERNAL_SERVER_ERROR,
   NO_CODEMOD_TO_PUBLISH,
   NO_CONFIG_FILE_FOUND,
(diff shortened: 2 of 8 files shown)


.changeset/three-bats-glow.md
```diff
@@ -0,0 +1,7 @@
+---
+"@codemod-com/backend": major
+"codemod": major
```
mohebifar: We should make this a `minor` or `patch` version upgrade. We also don't need to bump backend.


```suggestion
"codemod": patch
```

apps/backend/src/handlers/createAPIKeyHandler.ts
```diff
@@ -1,35 +1,17 @@
...
-  const apiKey = await createApiKey({
-    externalId: user.id,
-    apiKeyData: parse(createAPIKeyRequestSchema, request.body),
-    uuid,
+  Reply: ApiResponse<CreateAPIKeyResponse>;
+}> = async (request: UserDataPopulatedRequest, reply) => {
+  return reply.code(410).send({
+    errorText:
+      "This endpoint is no longer supported. Please use the new CLI instead.",
+    error: GONE,
```
mohebifar: I don't think we need to block any of the API key endpoints

apps/backend/src/publishHandler.ts
```diff
@@ -1,445 +1,20 @@
...
-  } finally {
-    await fs.promises.rm(unpackPath, {
-      recursive: true,
-      force: true,
-    });
-  }
+  return reply.code(410).send({
+    errorText:
+      "This endpoint is no longer supported. Please use the new CLI instead.",
+    error: GONE,
```
mohebifar: This is incorrect. We need to prevent publishing "new codemods". But releasing a new version for the "existing codemods" should be allowed. 

This is preventing any publish command.

apps/backend/src/unpublishHandler.ts
```diff
@@ -16,142 +8,9 @@ export type UnPublishHandlerResponse =
...
-  } catch (err) {
-    console.error(err);
-    return reply.code(500).send({
-      error: `Failed calling unpublish endpoint: ${(err as Error).message}`,
-      success: false,
-    });
-  }
+  return reply.code(410).send({
+    error:
+      "This endpoint is no longer supported. Please use the new CLI instead.",
```
mohebifar: Un-publishing should be allowed even encouraged 😁

apps/cli/src/main.ts
```diff
@@ -11,21 +11,8 @@ import {
...
-  handleCreateAPIKeyCommand,
-  handleDeleteAPIKeysCommand,
-  handleListAPIKeysCommand,
-} from "#commands/api-keys.js";
 import { handleFeedbackCommand } from "#commands/feedback.js";
-import { handleInitCliCommand } from "#commands/init.js";
-import { handleLearnCliCommand } from "#commands/learn.js";
-import { handleListNamesCommand } from "#commands/list.js";
-import { handleLoginCliCommand } from "#commands/login.js";
-import { handleLogoutCliCommand } from "#commands/logout.js";
```
mohebifar: We still need login, logout, publish, unpublish, whoami, list.

For the publish endpoint, when the endpoints returns `GONE` code, we'll need to show the deprecation warning

mohebifar (changes requested):

.changeset/three-bats-glow.md
amirabbas-gh: @mohebifar done

apps/backend/src/handlers/createAPIKeyHandler.ts
amirabbas-gh: @mohebifar fixed

apps/backend/src/publishHandler.ts
amirabbas-gh: @mohebifar done

apps/backend/src/unpublishHandler.ts
amirabbas-gh: @mohebifar done

apps/cli/src/main.ts
amirabbas-gh: @mohebifar done

apps/cli/src/main.ts
```diff
@@ -11,23 +11,16 @@ import {
 } from "@codemod-com/telemetry";
 import { doubleQuotify, execPromise } from "@codemod-com/utilities";
 import { version } from "#/../package.json";
-import {
```
mohebifar: We shouldn't remove the api key commands. These should continue to work.

mohebifar (changes requested):

apps/cli/src/main.ts
amirabbas-gh: @mohebifar ok, i got it, sorry for misunderstanding.
it's now almost fixed :)

mohebifar (approved):

PR #1603, opened 2025-07-13
chore: remove usless js dep

AugustinMauroy:
## Description 

- Remove useless dep
- only use `catalog` when it's a shared dep
- put dev/prod dep as correct place
- update ci to make it pass

> I may have forgotten something, but normally it's fine.


mohebifar:
Thanks so much! Did you use an automated tool for this?

btw, CI is failing.

AugustinMauroy:
> Thanks so much! Did you use an automated tool for this?

HEHE no 😥

PR #1609, opened 2025-07-14
fea(@codemod-com/codemod-standalone): explore

AugustinMauroy:
## Description 

As discussed with @alexbit-codemod it's may be useful to furnish a package that allow codemod writer to create both js  script and codemod jssg

So here a PoC. it's still need some test but yeah it's covered 80%

### DIFF
--- a/packages/codemod-standalone/README.md
+++ b/packages/codemod-standalone/README.md
@@ -1,0 +1,41 @@
+# Codemod Standalone
+
+## Usage
+
+```ts
+import { registerCodemod } from "@codemod-com/codemod-standalone";
+import type { SgRoot } from "@ast-grep/napi";
+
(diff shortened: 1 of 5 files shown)


alexbit-codemod:
thank you for the amazing work. you rock. let @mohebifar review it once he has some bandwidth.

mohebifar:
This is amazing and aligned with what we were envisioning too. It's a bit hard to maintain parity between the two systems, but I can see the value in having the option to use via nodejs.

AugustinMauroy:
> but I can see the value in having the option to use via nodejs.

I have to check but I think, it's can also run with deno & bun

AugustinMauroy:
bump @mohebifar

PR #1617, opened 2025-07-16
chore: Update README

mohab-sameh:
Updates repo readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,59 +16,35 @@
   </p>
 </p>
 
-## Codemod
+# Codemod
 
-Codemod platform helps you create, distribute, and run codemods in codebases of any size.
+[![Docs](https://img.shields.io/badge/docs-available-brightgreen)](https://docs.codemod.com)
+[![Community](https://img.shields.io/badge/slack-join-blue)](https://codemod.com/community)
+[![License](https://img.shields.io/github/license/codemod-com/codemod)](LICENSE)
 
-The AI-powered, community-led codemods enable you to automate framework upgrades, large refactoring, and boilerplate programming with unparalleled speed and developer experience.
+Codemod is the AI-powered, community-led platform for automating code migrations, framework upgrades, and large-scale changes.
 
-We are making dream migrations a reality for all developers in the world, one framework at a time.
+## Get Started
 
-## Documentation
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,59 +16,35 @@
...
 
-The AI-powered, community-led codemods enable you to automate framework upgrades, large refactoring, and boilerplate programming with unparalleled speed and developer experience.
+Codemod is the AI-powered, community-led platform for automating code migrations, framework upgrades, and large-scale changes.
 
-We are making dream migrations a reality for all developers in the world, one framework at a time.
+## Get Started
 
-## Documentation
+- **Platform:** [Codemod Platform](https://go.codemod.com/app) — Plan, automate, and track migrations in any codebase.
+- **Docs:** [docs.codemod.com](https://docs.codemod.com)
```
AugustinMauroy, line 29: add new registry link

README.md
```diff
@@ -16,59 +16,35 @@
   </p>
 </p>
 
-## Codemod
+# Codemod
 
-Codemod platform helps you create, distribute, and run codemods in codebases of any size.
+[![Docs](https://img.shields.io/badge/docs-available-brightgreen)](https://docs.codemod.com)
+[![Community](https://img.shields.io/badge/slack-join-blue)](https://codemod.com/community)
```
AugustinMauroy:
```suggestion
[![Community](https://img.shields.io/badge/slack-join-brightgreen)](https://codemod.com/community)
```

keep it green like the rest

README.md
mohab-sameh: The new registry is still shadow-published. @mohebifar You still don't want to make it the default one yet, correct?

AugustinMauroy: > The new registry is still shadow-published

Yeah but codemod you write on studio is disigned to be publish on the new registry so that make sense to allow user to see it

README.md
mohab-sameh: changed to slack's yellow color. contrast in readme badges is common and looks better.

README.md
mohab-sameh: I think it's a bit of a transitional phase we're in. The new registry in its current state cannot replace the legacy one yet. That would even have significant negative SEO impact. However, it's important this phase doesn't last for long and all existing codemods are transferred to the new registry. cc @mohebifar @alexbit-codemod

PR #1618, opened 2025-07-16
feat(CLI): add git dirty check for `run` commands

amirabbas-gh:
#### 📚 Description

This PR adds support for detecting a dirty Git working tree before executing codemod using run commands. This feature has been integrated into the following commands:
- `run`
- `jssg/run`
- `workflow/run`

If the repository contains uncommitted changes, a warning will be displayed to inform users before proceeding. This ensures improved awareness and safety during codemod executions.

#### 🧪 Test Plan
- Verified that the dirty state check works for all 3 run-related commands.
- Manually tested the behavior with both clean and dirty Git states.
- Confirmed that warnings are shown appropriately and command behavior remains unchanged otherwise.
- Confirmed `--allow-dirty` is working as well

### DIFF
--- a/crates/cli/src/dirty_git_check.rs
+++ b/crates/cli/src/dirty_git_check.rs
@@ -1,10 +1,14 @@
 use anyhow::Result;
 use inquire::Confirm;
-use std::process::Command as ProcessCommand;
+use std::path::Path;
+use std::process::Command;
 
-pub fn dirt_check(allow_dirty: bool) -> Result<()> {
-    if !allow_dirty {
-        let output = ProcessCommand::new("git")
+pub fn dirty_check(allow_dirty: bool) -> Result<()> {
+    if !allow_dirty
+        && Command::new("git").arg("--version").output().is_ok()
+        && Path::new(".git").exists()
+    {
+        let output = Command::new("git")
             .args(["status", "--porcelain"])
             .output()
             .expect("Failed to run git");
--- a/crates/cli/src/dirty_git_check.rs
+++ b/crates/cli/src/dirty_git_check.rs
@@ -13,10 +13,7 @@ pub fn dirty_check(allow_dirty: bool) -> Result<()> {
             .output()
             .expect("Failed to run git");
 
-        if !output.stdout.is_empty()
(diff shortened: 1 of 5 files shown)


crates/cli/src/dirty_git_check.rs
```diff
@@ -0,0 +1,28 @@
...
+                .contains("nothing to commit, working tree clean")
+        {
+            print!("⚠️  You have uncommitted changes. Do you want to continue anyway? [y/N]: ");
+            io::stdout().flush().unwrap();
+            let mut answer = String::new();
+            io::stdin().read_line(&mut answer).unwrap();
+            if !matches!(answer.trim().to_lowercase().as_str(), "y" | "yes") {
+                return Err(anyhow::anyhow!("Aborting due to uncommitted changes"));
+            }
+        }
```
mohebifar: Use inquire::Confirm.

crates/cli/src/dirty_git_check.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use anyhow::Result;
+use std::process::Command as ProcessCommand;
+
+pub fn dirt_check(allow_dirty: bool) -> Result<()> {
+    if !allow_dirty {
+        let output = ProcessCommand::new("git")
+            .args(["status", "--porcelain"])
+            .output()
+            .expect("Failed to run git");
```
mohebifar: A few things:
- The user may not have `git` installed.
- The dir may not be a git dir (git is not initialized in that dir).

In these cases, we should ignore this check

mohebifar (changes requested):

crates/cli/src/dirty_git_check.rs
```diff
@@ -0,0 +1,33 @@
+use anyhow::Result;
+use inquire::Confirm;
+use std::path::Path;
+use std::process::Command;
+
+pub fn dirty_check(allow_dirty: bool) -> Result<()> {
+    if !allow_dirty
+        && Command::new("git").arg("--version").output().is_ok()
```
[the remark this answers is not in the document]
mohebifar: Nah, it's fine. Ignore

PR #1620, opened 2025-07-18
Feat(jssg): better validation

AugustinMauroy:
## Description

It's reduce the amount of possible bug 

- validate js entry point if they exist
- clean serval thing


crates/core/src/utils.rs
```diff
@@ -1,14 +1,20 @@
+use anyhow::{Context, Result};
```
mohebifar, line 1: We use anyhow in CLI only. In CLI, we use anyhow for ergonomic error propagation and user-facing reports.
In core/library crates, we define typed errors with `thiserror`.

mohebifar (commented):
CI is failing. Have you tried running `cargo check` and `cargo clippy --fix --allow-dirty && cargo fmt`?

crates/core/src/utils.rs
AugustinMauroy: got it I had moved the function from cli to core to reused it so I have to modify that

PR #1633, opened 2025-07-22
chore(template): update

AugustinMauroy:
## Description 

Update JSSG example to make it functional and related to lasted change

### DIFF
--- a/crates/cli/src/templates/js-astgrep/package.json
+++ b/crates/cli/src/templates/js-astgrep/package.json
@@ -4,14 +4,14 @@
   "description": "{description}",
   "type": "module",
   "dependencies": {
-    "@ast-grep/cli": "^0.15.0",
-    "@ast-grep/napi": "^0.15.0"
+    "@ast-grep/napi": "^0.39.1"
   },
   "devDependencies": {
-    "@codemod.com/jssg-types": "^1.0.3"
+    "@codemod.com/jssg-types": "^1.0.3",
+    "typescript": "^5.8.3"
   },
   "scripts": {
-    "test": "node tests/test.js",
-    "apply": "node scripts/apply-codemod.js"
+    "test": "npx codemod@next jssg test -l typescript ./scripts/codemod.js.ts",
+    "check-types": "npx tsc --noEmit"
   }
 }
--- a/crates/cli/src/templates/js-astgrep/workflow.yaml
+++ b/crates/cli/src/templates/js-astgrep/workflow.yaml
@@ -1,4 +1,5 @@
 version: "1"
+
 nodes:
   - id: install-deps
     name: Install Dependencies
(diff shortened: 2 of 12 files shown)


mohab-sameh (approved):

crates/cli/src/templates/js-astgrep/package.json
```diff
@@ -4,14 +4,14 @@
...
   },
   "devDependencies": {
-    "@codemod.com/jssg-types": "^1.0.3"
+    "@codemod.com/jssg-types": "^1.0.3",
+    "typescript": "^5.8.3"
   },
   "scripts": {
-    "test": "node tests/test.js",
-    "apply": "node scripts/apply-codemod.js"
+    "test": "npx codemod@next jssg test -l typescript ./scripts/codemod.js.ts",
```
mohebifar:
```suggestion
    "test": "npx codemod@next jssg test -l typescript ./scripts/codemod.ts",
```

crates/cli/src/templates/js-astgrep/workflow.yaml
```diff
@@ -17,13 +18,13 @@ nodes:
     runtime:
       type: direct
     steps:
-      - name: "Remove console.log statements"
+      - name: Remove console.log statements
         js-ast-grep:
-          js_file: "scripts/codemod.ts"
-          base_path: "."
+          js_file: scripts/codemod.js.ts
```
mohebifar: Why `.js.ts`?

mohebifar (commented):
Looks great! And this was so overdue! Thanks. Just a comment on the codemod file name

crates/cli/src/templates/js-astgrep/package.json
```diff
@@ -4,14 +4,14 @@
...
   "type": "module",
   "dependencies": {
-    "@ast-grep/cli": "^0.15.0",
-    "@ast-grep/napi": "^0.15.0"
+    "@ast-grep/napi": "^0.39.1"
   },
   "devDependencies": {
-    "@codemod.com/jssg-types": "^1.0.3"
+    "@codemod.com/jssg-types": "^1.0.3",
+    "typescript": "^5.8.3"
```
mohebifar, line 11: Good call!

crates/cli/src/templates/js-astgrep/workflow.yaml
AugustinMauroy: because in the example the transform function that touch js file is call like that

mohebifar: Ah, that's the template file names. When the file gets created, it's actually named `codemod.ts` always: https://github.com/codemod-com/codemod/blob/3223123fad77b83b7c51b05a353d9ed7660fccbf/crates/cli/src/commands/init.rs#L436

that `.<lang>` extension prefix is just for the template file name in our source code.

AugustinMauroy: cool cool cool let's me update that

PR #1639, opened 2025-07-22
feat(cli): bundle jssg codemods pre-publish

mohebifar:
#### 📚 Description
- Introduced a new `bundle` command in the JSSG CLI for bundling JSSG codemods and their dependencies.
- Implemented a bundler utility that processes and bundles JS files, integrating with existing module resolution.
- Refactored existing commands to accommodate the new bundling functionality.
- Auto bundle JSSG codemods on publish

### DIFF
--- a/crates/codemod-sandbox/src/utils/bundler/mod.rs
+++ b/crates/codemod-sandbox/src/utils/bundler/mod.rs
@@ -1,0 +1,593 @@
+use crate::sandbox::resolvers::{oxc_resolver::OxcResolver, traits::ModuleResolver};
+use crate::utils::bundler::module_analyzer::{ModuleAnalyzer, ModuleExports};
+use crate::utils::bundler::module_transformer::ModuleTransformer;
+use crate::utils::transpiler;
+use std::sync::Arc;
+use std::{
+    collections::{HashMap, HashSet},
+    path::PathBuf,
+};
+use swc_core::common::SourceMap;
+use swc_core::ecma::codegen::{text_writer::JsWriter, Emitter};
+use swc_core::{
+    common::{BytePos, FileName},
+    ecma::{
+        ast::{EsVersion, ModuleItem},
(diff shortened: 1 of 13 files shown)


crates/codemod-sandbox/src/utils/bundler/mod.rs
```diff
@@ -0,0 +1,593 @@
...
+}
+
+/// Runtime module system options
+#[derive(Debug, Clone)]
+pub enum RuntimeSystem {
+    /// CommonJS-style module system
+    CommonJS,
+    /// Custom runtime with require/module.exports
+    Custom,
+}
```
AugustinMauroy, line 64: CJS is actually  `require`/`module.exports` 
and esm is `import`/`export`

mohebifar: It just means that the code in the module factory turns into CJS style code i.e. `import {X} from 'module'` is converted to `const {X} = __codemod_require('module')` and `export { X }` is converted to `exports.X = X`.

Under the hood, no import or export actually happens as the modules are stored in an object record.

mohebifar:
> maybe during bundling process. You can check if import is `node:*` is codemod have this api let's use rewite it instead throw an error

Ah good point to account for `node:*`. Will fix in a follow-up PR. This has to be fixed on both

AugustinMauroy:
> Ah good point to account for `node:*`. Will fix now.

Verify if `jssg` correctly run `node:*` instead you I'll need to review it before run or make an alias.

crates/codemod-sandbox/src/utils/bundler/mod.rs
AugustinMauroy: make sens maybe rephrase little bit the comment

mohebifar:
It just works. LLRT already comes with aliases for `node:*`. So all is good as per my tests.

PR #1651, opened 2025-07-24
chore(docs): Changelog update July 24

mohab-sameh:
Changelog update for July 24, 2025

Screenshot 2025-07-24 at 8 39 05 PM

### DIFF
--- a/apps/docs/more-resources/changelog.mdx
+++ b/apps/docs/more-resources/changelog.mdx
@@ -5,7 +5,19 @@ icon: 'scroll'
 mode: "center"
 ---
 
-<Update label="2025-06-03">
+<Update label="July 24, 2025">
+
+- **API Keys Management**: Generate, rotate, and revoke keys with per-package permissions and automatic expiry.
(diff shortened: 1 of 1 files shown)


mohebifar:
Let's remove:
- create workflows from registry (It's behind a feature flag)
- security hardening: we don't need to mention the details perhaps? Just give some generic context.
- performance & build: customers don't care about turbopack and nextjs version :D

apps/docs/more-resources/changelog.mdx
```diff
@@ -5,7 +5,18 @@ icon: 'scroll'
 mode: "center"
 ---
 
-<Update label="2025-06-03">
+<Update label="July 24, 2025">
+
+- **API Keys Management**: Generate, rotate, and revoke keys with per-package permissions and automatic expiry.
```
mohebifar: it's not per-package (yet). it's per-user/org.

AugustinMauroy:
maybe include upgrade on CLI with JS bundling which is a great DX improvement

PR #1655, opened 2025-07-27
fix(codemod-sandbox): Replace row with line in JsPosition to match the types from @ast-grep/napi

brunocroh:
### 📚 Description
The type `Position` in @codemod.com/jssg-types differs from what is returned in Range objects in JavaScript.

Looking at the ast-grep Rust implementation, the line is returned as row there. If this change needs to be made in the Rust code rather than in the type definitions, please let me know, I’d be happy to contribute to the Rust code (I tried to run it but had some issues setting up the development environment, so I’ll need some help with it. I can also use this opportunity to document how to setup development environment in the CONTRIBUTING.md).

https://github.com/codemod-com/codemod/blob/6578c1672bde74559c341e8adf89556c7bc464c5/crates/codemod-sandbox/src/ast_grep/sg_node.rs#L154-L165

#### 🔗 Linked Issue
https://github.com/nodejs/userland-migrations/pull/136

### DIFF
--- a/crates/codemod-sandbox/src/ast_grep/sg_node.rs
+++ b/crates/codemod-sandbox/src/ast_grep/sg_node.rs
@@ -153,12 +153,12 @@ impl<'js> SgNodeRjs<'js> {
 
         let result = JsNodeRange {
             start: crate::ast_grep::types::JsPosition {
-                row: start_pos_obj.line(),
+                line: start_pos_obj.line(),
                 column: start_pos_obj.column(&self.inner_node),
                 index: byte_range.start,
             },
             end: crate::ast_grep::types::JsPosition {
-                row: end_pos_obj.line(),
+                line: end_pos_obj.line(),
                 column: end_pos_obj.column(&self.inner_node),
                 index: byte_range.end,
             },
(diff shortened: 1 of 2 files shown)


brunocroh:
> Thanks for pointing that out! Good catch. I believe the change should be made in the Rust code instead. Let me know if you’d like to update the PR or if you need any assistance.

Okay, so I dropped the previous commit and pushed a new one. Now I'm updating the JsPosition object to replace row with line.
If I understand the code correctly, this is the bridge between wasm and Rust, so I'm just updating the names on the bridge/JavaScript side — the sandbox logic remains unchanged.

Let me know if anything else needs to be done, thank you.

brunocroh:
> Thanks for pointing that out! Good catch. I believe the change should be made in the Rust code instead. Let me know if you’d like to update the PR or if you need any assistance.

> Thanks for pointing that out! Good catch. I believe the change should be made in the Rust code instead. Let me know if you’d like to update the PR or if you need any assistance.

Okay, so I dropped the previous commit and pushed a new one. Now I'm updating the JsPosition object to replace row with line.
If I understand the code correctly, this is the bridge between wasm and Rust, so I'm just updating the names on the bridge/JavaScript side, the sandbox logic remains unchanged.

Let me know if anything else needs to be done, thank you.

mohebifar:
Thank you! I tested these changes and it all looks good. Waiting for CI 👀

mohebifar (approved):

mohebifar:
Thank you so much! These changes will be included in the next release which would be `1.0.0-rc.19`

brunocroh:
Brilliant, thank you mate!

PR #1663, opened 2025-07-27
fix: prevent publishing codemods when jssg module resolution fails

mohebifar:
#### 📚 Description

Previously, the _bundler errors_ on publish were getting swallowed. We should stop the publish process when jssg entry file cannot be bundled.

### DIFF
--- a/crates/codemod-sandbox/src/sandbox/errors.rs
+++ b/crates/codemod-sandbox/src/sandbox/errors.rs
@@ -18,7 +18,7 @@ pub enum FsError {
 
 #[derive(Debug, Error)]
 pub enum ResolverError {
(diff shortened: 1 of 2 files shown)


crates/codemod-sandbox/src/sandbox/errors.rs
```diff
@@ -18,7 +18,7 @@ pub enum FsError {
 
 #[derive(Debug, Error)]
 pub enum ResolverError {
-    #[error("Module resolution failed: base='{base}', name='{name}'")]
+    #[error("Module resolution failed:\nCannot resolve {name} from {base}")]
```
mohebifar, line 21: I think it's more clear and aligned with node.js module resolution error.

PR #1678, opened 2025-08-04
fix(website): Newsletter form token error

mohab-sameh:
### 🛠  What’s fixed
Users occasionally saw “Captcha token missing” because the browser submitted the form before `grecaptcha` was ready.  
This patch guarantees a token is generated (or the user gets a clear error) before any request is sent.

### ✨  Key changes
1. **apps/frontend/hooks/useFormSubmission.tsx**
   • Added `waitForRecaptcha()` (max-wait 5 s) to ensure the API is loaded.  
   • If reCAPTCHA can’t load or execute, the form shows a friendly error and aborts.  
   • On success, token is appended exactly as before.

2. **No server-side changes** – API route logic remains the same.

### ✅  Outcome
• Eliminates “Captcha token missing” errors on the backend.  
• Users see actionable feedback when reCAPTCHA fails instead of a silent 400.

### DIFF
--- a/apps/frontend/hooks/useFormSubmission.tsx
+++ b/apps/frontend/hooks/useFormSubmission.tsx
@@ -30,6 +30,23 @@ export function useFormSubmission(options: { recaptcha?: boolean } = {}) {
     }
   }, [options.recaptcha, siteKey]);
 
+  async function waitForRecaptcha(): Promise<any | null> {
+    if ((window as any).grecaptcha) return (window as any).grecaptcha;
+    return new Promise((resolve) => {
+      const maxWait = 5000;
+      const start = Date.now();
+      const interval = setInterval(() => {
+        if ((window as any).grecaptcha) {
+          clearInterval(interval);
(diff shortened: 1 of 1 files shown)


alexbit-codemod (approved):

apps/frontend/next.config.mjs
```diff
@@ -63,6 +63,25 @@ const config = {
...
+              "script-src 'self' https://www.google.com/recaptcha/ https://www.gstatic.com/recaptcha/ 'unsafe-inline' 'unsafe-eval'; " +
+              "frame-src https://www.google.com/recaptcha/ https://www.gstatic.com/recaptcha/; " +
+              "style-src 'self' 'unsafe-inline'; " +
+              "img-src * data: blob:; " +
+              "connect-src *;",
+          },
+        ],
+      },
+    ];
+  },
```
mohebifar, line 84: Let's revert this @mohab-sameh

PR #1686, opened 2025-08-08
(draft) Progress bar for run

amirabbas-gh:


crates/cli/Cargo.toml
```diff
@@ -16,7 +16,7 @@ butterflow-state = { workspace = true }
 butterflow-runners = { workspace = true }
 codemod-sandbox = { workspace = true }
 codemod-telemetry = { workspace = true }
-tokio = { workspace = true }
+tokio.workspace = true
```
mohebifar: ?

crates/core/src/registry.rs
```diff
@@ -333,91 +347,160 @@ impl RegistryClient {
...
+                            .unwrap_or(reqwest::StatusCode::INTERNAL_SERVER_ERROR)
+                            .into(),
+                        message: e.to_string(),
+                    })?;
+
+                let total_size = head_response
+                    .headers()
+                    .get(CONTENT_LENGTH)
+                    .and_then(|val| val.to_str().ok()?.parse().ok())
+                    .unwrap_or(0);
```
mohebifar: should we default to a non-zero value like 1 so we don't run into a division by zero error?

crates/core/src/registry.rs
```diff
@@ -333,91 +347,160 @@ impl RegistryClient {
...
+                                                    actual_magic[0], actual_magic[1]
+                                                ),
+                                            });
+                                        }
+                                    }
+
+                                    // Extract the actual package data
+                                    self.extract_package(&package_data, cache_dir).await?;
+                                    info!("Package cached to: {}", cache_dir.display());
+                                    return Ok(cache_dir.to_path_buf());
```
mohebifar: I'll add a todo to clean this up later. This looks verbose.

crates/codemod-sandbox/src/sandbox/engine/execution_engine.rs
```diff
@@ -287,51 +302,7 @@ where
...
-                                                    source.to_string()
-                                                }
-                                                _ => e.to_string(),
-                                            }
-                                        );
-                                        errors.lock().unwrap().push(error_msg);
-                                    }
-                                }
-                            });
+                            final_paths.lock().unwrap().push(file_path.to_path_buf());
```
mohebifar: this effectively eliminates the benefits of multi-threading. the locking mechanism restricts processing to one file at a time. My benchmark proves this:

```
BEFORE: 180.23s user 4.49s system 1169% cpu 15.789 total
AFTER: 142.05s user 4.30s system 98% cpu 2:27.88 total
```

mohebifar (changes requested):
We don't really need to show the current file name if it requires locking. This current approach is eliminating parallelism.

PR #1693, opened 2025-08-11
Fix/add registry redirect

siamak:


apps/frontend/next.config.mjs
```diff
@@ -116,6 +116,11 @@ const config = {
         destination: "/registry/:slug*",
         permanent: true,
       },
+      {
+        source: "/registry/:path*",
+        destination: "https://app.codemod.com/registry/:path*",
+        permanent: true,
```
mohebifar, line 122: @mohab-sameh is permanent redirect required? maybe we can start with 302 before moving to 301.

PR #1695, opened 2025-08-12
fix: fix git dirty condition

amirabbas-gh:


mohebifar:
I incorporated these changes in #1696

PR #1715, opened 2025-08-20
fix(cli): add shorthand for --update-snapshot arg

brunocroh:
#### 📚 Description

The [docs for the command](https://docs.codemod.com/cli/cli-reference#param-update-snapshots-u) `codemod jssg test` specify that the argument `--update-snapshots` has the shorthand `-u`.
However, it doesn’t actually exist, so this PR adds the shorthand.


mohab-sameh:
@all-contributors add @brunocroh for code

PR #1730, opened 2025-09-01
Feat: add capabilities feature

amirabbas-gh:
#### 📚 Description

# Draft

### DIFF
--- a/crates/cli/src/capabilities_security_callback.rs
+++ b/crates/cli/src/capabilities_security_callback.rs
@@ -1,0 +1,33 @@
+use inquire::Confirm;
+use log::error;
+use std::sync::{Arc, Mutex};
+
+use butterflow_core::execution::CodemodExecutionConfig;
+
+type CapabilitiesSecurityCallback = Arc<Box<dyn Fn(&CodemodExecutionConfig) + Send + Sync>>;
+
+pub fn capabilities_security_callback() -> CapabilitiesSecurityCallback {
+    let checked_capabilities = Arc::new(Mutex::new(Vec::<String>::new()));
+
+    Arc::new(Box::new(move |config: &CodemodExecutionConfig| {
+        for capability in config.capabilities.as_ref().unwrap_or(&Vec::new()) {
+            let answer = Confirm::new(&format!(
+            "🛡️  \x1b[31mSecurity Notice\x1b[0m: This action will grant access to `{capability}`, which may perform sensitive operations. Are you sure you want to continue?"
+        ))
+        .with_default(false)
+        .with_help_message("Press 'y' to continue or 'n' to abort")
+        .prompt().unwrap_or_else(|e| {
+            error!("Failed to get user input: {e}");
+            std::process::exit(1);
+        });
+
+            if !answer {
+                error!("Aborting due to capabilities warning");
+                std::process::exit(1);
+            } else {
+                let mut checked_capabilities = checked_capabilities.lock().unwrap();
+                checked_capabilities.push(capability.to_string());
+            }
+        }
+    }))
+}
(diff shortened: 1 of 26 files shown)


crates/cli/src/capabilities_security_callback.rs
```diff
@@ -0,0 +1,33 @@
...
+            error!("Failed to get user input: {e}");
+            std::process::exit(1);
+        });
+
+            if !answer {
+                error!("Aborting due to capabilities warning");
+                std::process::exit(1);
+            } else {
+                let mut checked_capabilities = checked_capabilities.lock().unwrap();
+                checked_capabilities.push(capability.to_string());
```
mohebifar: This would display the prompt once for each capability. We should show one prompt only with all capabilities listed in that single message

crates/core/src/execution.rs
```diff
@@ -57,6 +57,8 @@ pub struct CodemodExecutionConfig {
     pub dry_run: bool,
     /// Language
     pub languages: Option<Vec<String>>,
+    /// Capabilities
+    pub capabilities: Option<Vec<String>>,
```
mohebifar: We should make capabilities an enum type or bit flags instead of `Vec<String>`

crates/cli/src/capabilities_security_callback.rs
```diff
@@ -0,0 +1,48 @@
...
+            std::process::exit(1);
+        });
+
+        let mut checked = checked_capabilities.lock().unwrap();
+        checked.extend(need_to_check);
+        drop(checked);
+
+        if !answer {
+            error!("Aborting due to capabilities warning");
+            std::process::exit(1);
```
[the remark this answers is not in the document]
mohebifar: Should we use anyhow here instead?

crates/cli/src/capabilities_security_callback.rs
```diff
@@ -0,0 +1,48 @@
...
+        let answer = Confirm::new(&format!(
+            "🛡️  \x1b[31mSecurity Notice\x1b[0m: This action will grant access to `{}`, which may perform sensitive operations. Are you sure you want to continue?", 
+            need_to_check.iter().map(|c| format!("{c:?}")).collect::<Vec<_>>().join(", ")
+        ))
+        .with_default(false)
+        .with_help_message("Press 'y' to continue or 'n' to abort")
+        .prompt().unwrap_or_else(|e| {
+            error!("Failed to get user input: {e}");
+            std::process::exit(1);
+        });
```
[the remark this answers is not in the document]
mohebifar: Should we use anyhow here instead?

crates/cli/src/capabilities_security_callback.rs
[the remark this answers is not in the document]
amirabbas-gh: @mohebifar done

crates/cli/src/capabilities_security_callback.rs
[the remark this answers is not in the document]
amirabbas-gh: @mohebifar done

crates/core/src/engine.rs
```diff
@@ -1458,11 +1460,37 @@ impl Engine {
...
+                                    "Invalid capability string '{}': {}",
+                                    v, e
+                                )));
+                            }
+                        }
+                    }
+                    Some(parsed_caps)
+                } else {
+                    None
+                };
```
mohebifar: @amirabbas-gh why are we using params here to construct llrt modules? 🤔

crates/llrt-capabilities/src/module_builder.rs
mohebifar, line 1: You can simplify this whole file using macros. e.g. `impl_safe_module!` and `impl_unsafe_module!`

crates/core/src/engine.rs
mohebifar: I was under the impression that we just use codemod.yaml as the source of truth for capability defs.

amirabbas-gh: @mohebifar fixed

mohebifar (approved):

PR #1780, opened 2025-09-18
feat: add repository choice into init

amirabbas-gh:
#### 📚 Description

Add git repository url prompt for init command

### DIFF
--- a/crates/cli/src/templates/codemod.yaml
+++ b/crates/cli/src/templates/codemod.yaml
@@ -6,7 +6,7 @@ description: "{description}"
 author: "{author}"
 license: "{license}"
 workflow: "workflow.yaml"
-repository: "{repository}"
+{repository}
 
 targets:
   languages: ["{language}"]
--- a/crates/cli/src/templates/codemod.yaml
+++ b/crates/cli/src/templates/codemod.yaml
@@ -6,6 +6,10 @@ description: "{description}"
 author: "{author}"
 license: "{license}"
 workflow: "workflow.yaml"
+repository: "{repository}"
+
+targets:
+  languages: ["{language}"]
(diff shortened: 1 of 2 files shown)


crates/cli/src/templates/codemod.yaml
```diff
@@ -6,6 +6,11 @@ description: "{description}"
 author: "{author}"
 license: "{license}"
 workflow: "workflow.yaml"
+category: "migration"
```
mohebifar: Let's remove `category`.

mohebifar: We should just simplify the default template. I don't have a good use-case for category yet.

amirabbas-gh: ok, sure

amirabbas-gh: ok sure

amirabbas-gh: @mohebifar done

crates/cli/src/templates/codemod.yaml
```diff
@@ -6,6 +6,10 @@ description: "{description}"
 author: "{author}"
 license: "{license}"
 workflow: "workflow.yaml"
+repository: "{repository}"
```
mohebifar: what happens if the user leaves the repository input blank? would it end up like this? `repository: ""`?

amirabbas-gh: yeah, it's end up like this: `repository: ""`
is it ok?

mohebifar: I'd say, if we just omit it, it'd be much better.

PR #1781, opened 2025-09-18
fix: fix disable analytic flag bug

amirabbas-gh:
#### 📚 Description

Things done in this PR:

- Converted verbose and disable analytics into a single global argument and fixed the issue with using `-‍‍-` before calling them
- Fixed the param issue in the jssg run and run commands
- Passed the `--no-telemetry` argument to the legacy command

### DIFF
--- a/crates/cli/src/commands/run.rs
+++ b/crates/cli/src/commands/run.rs
@@ -3,7 +3,7 @@ use crate::progress_bar::download_progress_bar;
 use crate::workflow_runner::run_workflow;
 use crate::TelemetrySenderMutex;
 use crate::CLI_VERSION;
-use anyhow::{Context, Result};
+use anyhow::Result;
 use butterflow_core::registry::RegistryError;
 use butterflow_core::utils::generate_execution_id;
 use butterflow_core::utils::parse_params;
@@ -35,8 +35,8 @@ pub struct Command {
     dry_run: bool,
 
     /// Additional arguments to pass to the codemod
-    #[arg(last = true)]
-    params: Vec<String>,
+    #[arg(long = "param", value_name = "KEY=VALUE")]
+    params: Option<Vec<String>>,
 
     /// Allow dirty git status
     #[arg(long)]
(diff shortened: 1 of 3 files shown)


crates/core/tests/utils_tests.rs
```diff
@@ -308,25 +308,25 @@ fn test_parse_params() {
...
+
+//     // Verify that parsing fails
+//     assert!(result.is_err());
+//     match result {
+//         Err(Error::Other(msg)) => {
+//             assert!(msg.contains("Invalid parameter format"));
+//         }
+//         _ => panic!("Expected Other error"),
+//     }
+// }
```
mohebifar: Why did you comment this out?

crates/core/src/utils.rs
```diff
@@ -196,9 +196,11 @@ pub fn parse_params(params: &[String]) -> Result<HashMap<String, String>> {
...
         let parts: Vec<&str> = param.splitn(2, '=').collect();
         if parts.len() != 2 {
-            return Err(Error::Other(format!(
-                "Invalid parameter format: {param}. Expected format: key=value"
-            )));
+            println!(
+                "\x1b[33mWarning\x1b[0m: Invalid parameter format: {}",
+                param
+            );
+            continue;
```
mohebifar: Is this change related? also we should use console::style for these ansi codes

crates/cli/src/commands/run.rs
```diff
@@ -35,8 +35,8 @@ pub struct Command {
     dry_run: bool,
 
     /// Additional arguments to pass to the codemod
-    #[arg(last = true)]
-    params: Vec<String>,
+    #[arg(long)]
+    param: Option<Vec<String>>,
```
mohebifar:
```suggestion
    #[arg(long = "param", value_name = "KEY=VALUE")]
    params: Vec<String>,
```

Also why do we need to make it Option? doesn't it yield an empty vector by default?

mohebifar (approved):

PR #1801, opened 2025-09-26
docs: improve tagging best practices guide

mohab-sameh:
Improves tagging guidelines and conventions in docs

### DIFF
--- a/apps/docs/cli/packages/package-structure.mdx
+++ b/apps/docs/cli/packages/package-structure.mdx
@@ -112,18 +112,20 @@ Example: `keywords: ["react", "v18-to-v19", "migration"]`
 
   **1. Transformation Type Tags**
 
-  Use one of these to describe **why** the codemod exists:
+  Consider these categories when describing **why** the codemod exists:
 
-  - **`breaking-change`** – adapts code to framework/library breaking API changes.
-  - **`feature-adoption`** – helps adoption of new optional or incremental features.
+  - **`upgrade`** – helps upgrade code to newer versions (encompasses both breaking changes and feature adoption).
+    You may also consider adding **one** of the following tags:
+    - **`breaking-change`** – adapts code to framework/library breaking API changes.
+    - **`feature-adoption`** – helps adoption of new optional or incremental features.
   - **`security`** – addresses known vulnerabilities or unsafe patterns.
-  - **`migration`** – helps migrate code from one version to another.
   - **`cross-migration`** – replaces one library/framework with another.
-  - **`best-practice-modernization`** – aligns code with community standards and modern idioms.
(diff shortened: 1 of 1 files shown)


apps/docs/cli/packages/package-structure.mdx
```diff
@@ -97,9 +97,82 @@ Languages targeted by this codemod (selected language during `codemod init`; edi
...
+  ---
+
+  **1. Transformation Type Tags**
+
+  Use one of these to describe **why** the codemod exists:
+
+  - **`breaking-change`** – adapts code to framework/library breaking API changes.
+  - **`feature-adoption`** – helps adoption of new optional or incremental features.
+  - **`security`** – addresses known vulnerabilities or unsafe patterns.
+  - **`migration`** – helps migrate code from one version to another.
```
alexbit-codemod: Breaking changes and feature adoption are more granular tags for upgrades. Migration is too broad and does not add value. Maybe we can have "upgrade" as a high level category for either feature adoption or breaking change. 

(Also, as a side note, building codemods that do both feature adoption and handle breaking change is not a good idea. Idk if this is a place to talk about these or not. Maybe not)

apps/docs/cli/packages/package-structure.mdx
```diff
@@ -97,9 +97,82 @@ Languages targeted by this codemod (selected language during `codemod init`; edi
...
+  - Keep tags concise (1–2 words).
+  - Use lowercase for consistency.
+  - Don’t overload with tags — 2–4 per codemod is ideal.
+  - Prioritize **transformation type + framework/library + version (if relevant)**.
+
+  ---
+
+  **1. Transformation Type Tags**
+
+  Use one of these to describe **why** the codemod exists:
```
alexbit-codemod: This needs to be rephrased. They don't have to. This list is just a recommendation and they can add other tags too.

apps/docs/cli/packages/package-structure.mdx
```diff
@@ -97,9 +97,82 @@ Languages targeted by this codemod (selected language during `codemod init`; edi
...
+
+  Here are some examples to illustrate how tags combine:
+
+  - **React Root API Upgrade (17 → 18)**
+      - Tags: `breaking-change`, `v17-to-v18`, `react`
+  - **Adopt React Hooks**
+      - Tags: `feature-adoption`, `react`
+  - **Migrate from Moment.js to Day.js**
+      - Tags: `cross-migration`, `momentjs`, `dayjs`
+  - **Remove Hardcoded Strings for i18n**
```
alexbit-codemod, line 172: Idk if best practice modernization is the best to describe i18n. I18n and a11y are big enough to deserve their own category. We should see i18n tag here. And maybe remove best practice tag?

PR #1810, opened 2025-09-30
document jssg

mohab-sameh:


apps/docs/guides/jssg/quickstart.mdx
```diff
@@ -0,0 +1,142 @@
+---
+title: "Quickstart"
+---
+
+Build a working JavaScript ast-grep (jssg) codemod in minutes.
+
+## What is jssg?
+
+JavaScript ast-grep (jssg) is a codemod framework that lets you write code transformations using JavaScript/TypeScript. Instead of complex AST manipulation, you use patterns to find and replace code structures.
```
mohebifar: I find the last sentence a bit confusing. It's as if TOS OS not ast based whereas it actually is

apps/docs/guides/jssg/quickstart.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+**How it works:**
+- Your transform runs in a **QuickJS** runtime with **AWS LLRT** module infrastructure
```
mohebifar: Let's say something like this:

jssg is a secure JavaScript runtime similar to Node.js with almost all of the built-in nodejs modules and globals e.g fs, process, etc

Then in perhaps a footnote-like "learn more" section we can say it's p powered by QuickJs for the JS engine and uses LLRT for node modules compatibility, and oxc for module resolution.

mohebifar: We should also say it comes with ast grep as a built in module too.

apps/docs/guides/jssg/quickstart.mdx
```diff
@@ -0,0 +1,142 @@
...
+- The `codemod:ast-grep` module exposes `SgRoot` and `SgNode` APIs for AST navigation
```
mohebifar: SgRoot and SgNode are pure types not actual values/classes.

So we soups make that clear.

The values that it exports are parse and parseAsync

apps/docs/guides/jssg/quickstart.mdx
```diff
@@ -0,0 +1,142 @@
...
+- You write pattern-based rules to find code, then create edits to transform it
```
mohebifar: You can write any ast-grep rules as JS objects (link to SgRule type in Api references)

apps/docs/guides/jssg/quickstart.mdx
```diff
@@ -0,0 +1,142 @@
...
+- Return a string with changes applied, or `null`/`undefined` to skip the file
```
mohebifar: It's kinda vague. We should say a jssg codemod module is a module that "export default"s a function that takes the SgRoot of a source file and returns the string of the modified source code or null/undefined...

apps/docs/guides/jssg/quickstart.mdx
```diff
@@ -0,0 +1,142 @@
...
+
+**Key concepts:**
+- **Patterns**: Use `console.log($ARG)` to match function calls with captured arguments
```
mohebifar: ??

apps/docs/guides/jssg/quickstart.mdx
```diff
@@ -0,0 +1,142 @@
...
+- **Rules**: Combine patterns with `any`, `all`, `kind`, and relational constraints
```
mohebifar: We should link to our ast-grep docs. It is the same as ast-grep rule configs

apps/docs/guides/jssg/quickstart.mdx
mohebifar: We're jumping right into a console log example in "key concepts"

mohab-sameh: Oops leftover while refactoring this section!

PR #1819, opened 2025-10-08
improve jssg and workflow docs

mohab-sameh:


apps/docs/jssg/advanced.mdx
```diff
@@ -0,0 +1,440 @@
...
+
+const transform: Transform<TSX> = async (root, options) => {
+  // Access matrix values passed from the workflow (if matrix strategy is used)
+  const team = (options.matrixValues as any)?.team;
+  const shardId = (options.matrixValues as any)?.shardId;
+
+  const filename = root.filename();
+
+  // fitsInShard is an example sharding predicate provided by your orchestrator or custom code
+  if (options.matrixValues && fitsInShard(filename, options.matrixValues)) {
```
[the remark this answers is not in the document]
mohab-sameh: @mohebifar how about something like the following?

```ts
import path from "path";
import { fitsInShard } from "codemodctl/sharding";
import type { Transform } from "codemod:ast-grep";
import type TSX from "codemod:ast-grep/langs/tsx";

const transform: Transform<TSX> = async (root, options) => {
  // Normalize for stable sharding across environments
  const filename = path.relative(process.cwd(), root.filename());

  // Only shard when orchestrator provides matrix context + file list
  const mv = options.matrixValues as { shard?: string; _meta_files?: string[] } | undefined;
  if (mv?.shard && mv?._meta_files) {
    if (!fitsInShard(filename, mv)) {
      return null; // Skip files not in this shard
    }
  }

  // ... apply edits for files that belong to this shard ...
  return null;
};

export default transform;
```

apps/docs/cli/cli-reference.mdx
```diff
@@ -81,45 +84,33 @@ npx codemod <package-name>
...
+  
+  ```bash
+  # Specify a target directory
+  npx codemod workflow run -w ./my-codemod -t /abs/path/to/repo
 
-# Specify a target directory
-npx codemod @codemod-com/my-test-pkg -t /abs/path/to/repo
-```
+  # Specify a target directory
+  npx codemod @codemod-com/my-test-pkg -t /abs/path/to/repo
```
alexbit-codemod, line 99: why are we using this outdated name here? @codemod-com/my-test-pkg? @mohab-sameh 
better to use some nodejs one.

apps/docs/cli/packages/quickstart.mdx
```diff
@@ -3,6 +3,10 @@ title: "Quickstart"
 description: "Create your first codemod package and run it locally."
 ---
 
+A **codemod package** is a set of files that define an automated code transformation. Codemod Packages let you create, test, and distribute reusable transformations that update source code across entire codebases.
```
alexbit-codemod, line 6: Codemod package

not codemod package (because codemod is a concept, and this case, the package is our, Codemod, opinionated way of packaging stuff.

and not Codemod Package. Package is just a concept that we dont care about giving it an identity. just Codemod package everywhere should be fine. @mohab-sameh

apps/docs/jssg/reference.mdx
```diff
@@ -723,27 +729,29 @@ export default function transform(root: any) {
...
+      return null;
+    }
+    
+    // Log each match for inspection
+    matches.forEach((match, i) => {
+      console.log(\`Match \${i}:\`, match.text());
+    });
+    
+    return rootNode.commitEdits(edits);
+    }
```
mohebifar, line 753: Is this formatted correctly? Should there be one more indent level?

mohab-sameh: There is some weird bug in mintlify here. It's rendering with extra indents.

PR #1833, opened 2025-10-23
fix(workflow): resolve base_path issue for run steps and add sudo support

amirabbas-gh:
#### 📚 Description

- Fixed an issue where run steps looked for shell files in the workflow’s execution directory instead of the workflow’s actual location.
- Now base_path correctly points to the workflow directory, preventing path resolution errors and making workflows with run steps easier to execute.
- Added sudo support for executing shell files when the command is run under sudo or as an administrator.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -674,6 +674,7 @@ dependencies = [
  "flate2",
  "futures-util",
  "ignore",
+ "libc",
  "log",
  "rand 0.9.2",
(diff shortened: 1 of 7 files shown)


mohebifar:
> Added sudo support for executing shell files when the command is run under sudo or as an administrator. 

Also I don't understand the use case for this. 

I think we can close this PR?

PR #1863, opened 2025-11-11
fix form

mohab-sameh:
fixes newsletter form security policy


alexbit-codemod:
i thinks its irrelevant now? closing

PR #1893, opened 2025-12-11
refactor: reafctor from text storing variables into Hashmap and add functioanlity

amirabbas-gh:
#### 📚 Description

Refactor text-based variable storage into a HashMap to improve performance under concurrent load, and add a getOrSetStepOutput function.


.changeset/nervous-hornets-reply.md
```diff
@@ -0,0 +1,5 @@
+---
```
mohebifar, line 1: @amirabbas-gh can you remove this file? we don't need use changeset for the `latest` channel. only legacy rn

PR #1934, opened 2026-01-07
doc: update OSS maintainer

alexbit-codemod:
removed unnecessary info, added logos for more credibility.

### DIFF
--- a/docs/oss.mdx
+++ b/docs/oss.mdx
@@ -201,14 +201,14 @@ import { MigrationsTemplateDemo } from "/snippets/migrations-template-demo.mdx";
 </Steps>
 
 <Tip>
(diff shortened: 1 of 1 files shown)


docs/oss.mdx
```diff
@@ -2,67 +2,177 @@
...
+        alt="React logo"
+        className="block opacity-100 dark:opacity-0" 
+        noZoom
+      />
+      <img 
+        src="./images/logos/react.svg" 
+        alt="React logo"
+        className="absolute inset-0 m-auto block opacity-0 dark:opacity-100" 
+        noZoom
+      />
```
mohebifar, line 48: why controlling visibility with opacity instead of block & hidden?

PR #1937, opened 2026-01-15
feat(cli): semantic testing

sahilmob:
#### 📚 Description
This pr adds the `--semantic` testing flag, which can be used when testing transforms that may produce semantically equivalent but textually different outputs

#### 🧪 Test Plan
- Tested locally
- Added a suite of tests for each language

Docs updated


docs/jssg/testing.mdx
```diff
@@ -49,6 +49,9 @@ npx codemod jssg test -l tsx ./scripts/codemod.ts -v
 
 # Run tests one at a time (helpful for debugging)
 npx codemod jssg test -l tsx ./scripts/codemod.ts --sequential
+
+# Use semantic comparison (ignores property ordering differences)
+npx codemod jssg test -l tsx ./scripts/codemod.ts --semantic
```
mohebifar: We use the word semantic in a different context e.g. flow and scope analysis. This can get confusing. I would just change this to

* `--strictness strict` which is the current behaviour that does string equality
* `--strictness cst` which means we compare the CSTs
* `--strictness ast` comparing the ast
* `--strictness loose` which means means comparing ast but skipping ordering of certain children.

sahilmob: addressed

crates/testing-utils/src/strictness/compare.rs
```diff
@@ -0,0 +1,1158 @@
...
+    F: FnOnce() -> Result<bool, CompareError>,
+{
+    try_compare().unwrap_or_else(|e| {
+        eprintln!(
+            "Warning: {} comparison failed ({}), falling back to exact string comparison",
+            mode_name, e
+        );
+        expected == actual
+    })
+}
```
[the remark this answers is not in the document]
mohebifar: @sahilmob I think falling back to `ast` would make sense as well.

PR #1950, opened 2026-01-28
fix: avoid sending failure kind of event when cdmd succeeds

AmirSa12:
I removed the failure kind of event. let me know if I'm missing something.

### DIFF
--- a/crates/cli/src/commands/run.rs
+++ b/crates/cli/src/commands/run.rs
@@ -81,6 +81,32 @@ pub struct Command {
     no_color: bool,
 }
 
+async fn send_failure_event(
+    telemetry: &TelemetrySenderMutex,
+    codemod_name: &str,
+    error_message: &str,
+) {
+    telemetry
+        .send_event(
+            BaseEvent {
+                kind: "failedToExecuteCommand".to_string(),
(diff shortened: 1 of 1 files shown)


sahilmob (approved):

crates/cli/src/commands/run.rs
```diff
@@ -183,25 +183,6 @@ pub async fn handler(
...
-                        "commandName".to_string(),
-                        "codemod.executeCodemod".to_string(),
-                    ),
-                    ("os".to_string(), std::env::consts::OS.to_string()),
-                    ("arch".to_string(), std::env::consts::ARCH.to_string()),
-                ]),
-            },
-            None,
-        )
-        .await;
```
mohebifar, line 203: wait, what?!

mohebifar (commented):
the code snippet you removed is the only place emitting `failedToExecuteCommand`. so that leaves no other calls to telemetry for the failure event. can you make sure we still send the failure event as well? 

We should send that event when:
- Downloading codemod fails
- During running the workflow:
  - A jssg or ast-grep step fails with all files "erroring"
  - A shell script step exits with an error code

PR #2071, opened 2026-04-07
chore(mcp): MCP improvments

mohab-sameh:
This PR tightens the lean Codemod AI/MCP setup by improving authoring guidance and package generation quality without reintroducing the old bulky local-doc architecture.

## What changed

- Keeps Codemod MCP guidance while adding small hot-path reminders for:
  - preserving the scaffold-selected package manager in generated package scripts and README commands
  - using `run:` instead of `command:` for shell workflow steps
  - re-checking the full package surface before stopping (`README.md`, `codemod.yaml`, `workflow.yaml`, `package.json`, tests/fixtures, renamed refs)
- Improves `codemod init` so generated README/next-step commands respect the selected package manager instead of defaulting back to `npm`/`npx`
- Extends `validate_codemod_package` to flag package-manager drift in `package.json` scripts

## Why

In evaluation runs and peer trials, generated codemods sometimes:
- rewrote `yarn`/`pnpm` packages to `npx`/`npm`
- used `command:` instead of the workflow schema’s `run:`
- updated one part of the package but forgot related files like README or workflow/config metadata

These changes are intended to prevent those failures with minimal, targeted guidance and one concrete validation check, while keeping the overall install/MCP architecture close to `main`.


crates/mcp/src/handlers/package_validation.rs
```diff
@@ -0,0 +1,925 @@
...
+            McpError::internal_error(format!("Failed to serialize response: {error}"), None)
+        })?;
+
+        Ok(CallToolResult::success(vec![Content::text(content)]))
+    }
+
+    async fn validate_package(
+        &self,
+        request: ValidateCodemodPackageRequest,
+    ) -> Result<ValidateCodemodPackageResponse> {
```
mohebifar, line 291: Every tool that we add comes with a context overhead cost. What is the justification for this?

crates/mcp/src/handlers/package_scaffold.rs
```diff
@@ -0,0 +1,449 @@
...
+
+#[derive(Clone)]
+pub struct PackageScaffoldHandler;
+
+impl PackageScaffoldHandler {
+    pub fn new() -> Self {
+        Self
+    }
+
+    #[tool(
```
mohebifar: i don't think we need this tool. the ai can just run `codemod init`. Instead, we should just simplify the `codemod init` command itself to not require license, author, and git URL via options. that should be enough.

crates/mcp/src/lib.rs
```diff
@@ -6,16 +6,202 @@ use serde_json::json;
...
+    title: &str,
+    sections: &[String],
+    fallback: &str,
+) -> String {
+    if sections.is_empty() {
+        return fallback.to_string();
+    }
+
+    format!(
+        "# {title}\n\nThese instructions are sourced from the public Codemod docs deployment (`docs.codemod.com`). Prefer this content over local fallback notes when they differ.\n\n{}\n",
```
mohebifar: Does the AI care? It doesn't see the fallback

crates/mcp/src/lib.rs
```diff
@@ -6,16 +6,202 @@ use serde_json::json;
...
+    supplement_title: &str,
+    supplement: &str,
+    fallback: &str,
+) -> String {
+    if sections.is_empty() {
+        return fallback.to_string();
+    }
+
+    format!(
+        "# {title}\n\n## {supplement_title}\n\n{supplement}\n\n---\n\nThese instructions are sourced from the public Codemod docs deployment (`docs.codemod.com`). Prefer this content over local fallback notes when they differ.\n\n{}\n",
```
mohebifar: same

crates/mcp/src/lib.rs
```diff
@@ -6,16 +6,202 @@ use serde_json::json;
...
+async fn build_public_docs_bundle(title: &str, urls: &[&str], fallback: &str) -> String {
+    let sections = fetch_public_doc_sections(urls).await;
+    build_public_docs_bundle_from_sections(title, &sections, fallback)
+}
+
+fn build_public_docs_bundle_with_supplement_from_sections(
+    title: &str,
+    sections: &[String],
+    supplement_title: &str,
+    supplement: &str,
```
mohebifar, line 273: this is not used. why do we need this separate fn?

docs/model-context-protocol.mdx
```diff
@@ -50,14 +50,22 @@ This is the main setup path for the AI-native CLI experience. It installs Codemo
...
+- **`get_ast_grep_gotchas`** - Get the highest-priority ast-grep gotchas before implementing transforms
+- **`scaffold_codemod_package`** - Scaffold a codemod package via the real `codemod init` CLI
+- **`validate_codemod_package`** - Validate whether a codemod package is still a starter scaffold or incomplete
+- **`get_jssg_instructions`** - Get docs-backed JSSG guidance for creating and testing transformation scripts
+- **`get_jssg_utils_instructions`** - Get docs-backed import utility guidance for `getImport`, `addImport`, and `removeImport`
+- **`get_jssg_runtime_capabilities`** - Get runtime and capability guidance for LLRT/Node APIs and multi-file JSSG work
+- **`get_codemod_cli_instructions`** - Get docs-backed CLI, package, and workflow guidance
+- **`get_codemod_creation_workflow`** - Get docs-backed codemod authoring workflow guidance with a small local supplement
+- **`get_codemod_troubleshooting`** - Get troubleshooting guidance for common Codemod CLI failures
+- **`get_codemod_maintainer_monorepo`** - Get maintainer guidance for codemod monorepos and version-hop workspaces
```
mohebifar: These are not tools anymore. these are resources.

crates/mcp/src/lib.rs
```diff
@@ -151,21 +408,27 @@ impl CodemodMcpServer {
...
+        let instructions_content = build_public_docs_bundle(
+            "Canonical Codemod CLI and Workflow Documentation",
+            &[
+                CLI_DOC_URL,
+                PACKAGE_STRUCTURE_DOC_URL,
+                WORKFLOW_REFERENCE_DOC_URL,
+            ],
+            CODEMOD_CLI_INSTRUCTIONS,
+        )
+        .await;
```
mohebifar: i think we should now remove these document tools. all agents now support MCP resources. so no need to have these massive docs duplicated as tools as well.

mohebifar (changes requested):

crates/mcp/src/lib.rs
mohab-sameh: @mohebifar `build_public_docs_bundle_with_supplement_from_sections` is used directly by the async wrapper `build_public_docs_bundle_with_supplement` in lib.rs. It's not super necessary but it helps separate fetch vs render

docs/model-context-protocol.mdx
mohab-sameh: @mohebifar checked this, and these are still tools in the current implementation. they are also exposed as resources.

both are currently true:
- they are declared as `#[tool(...)]`
- they are also returned from `list_resources` / `read_resource`

that said, I think the instinct behind your comment is reasonable. These docs-based guidance entrypoints are closer to resources conceptually, and we may want to move toward (in the future? or now?):
- docs/guidance = resources
- actions = tools

but we have not made that transition yet. the installed skill/command flow still tells agents to call these as tools.

tldr;
- I do not think the current wording is factually wrong
- but I do think we could rephrase it to something like “docs-backed guidance entrypoints available as tools and resources” if we want to make that clearer
- and if we want them to stop being tools, that should be an intentional follow-up change, not something we assume is already true

sahilmob (commented):
@mohab-sameh 

This branch doesn't compile

```shell
he name `package_manager_test_command` is defined multiple times
    --> crates/cli/src/commands/init.rs:1812:1
     |
1369 | fn package_manager_test_command(package_manager: Option<&str>) -> String {
     | ------------------------------------------------------------------------ previous definition of the value `package_manager_test_command` here
...
1812 | fn package_manager_test_command(package_manager: &str) -> String {
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `package_manager_test_command` redefined here
     |
     = note: `package_manager_test_command` must be defined only once in the value namespace of this module

error[E0308]: mismatched types
    --> crates/cli/src/commands/init.rs:1332:68
     |
1332 | ...AstGrepJs => package_manager_test_command(package_manager),
     |                 ---------------------------- ^^^^^^^^^^^^^^^ expected `Option<&str>`, found `&str`
     |                 |
     |                 arguments to this function are incorrect
     |
     = note:   expected enum `std::option::Option<&str>`
             found reference `&str`
note: function defined here
    --> crates/cli/src/commands/init.rs:1369:4
     |
1369 | fn package_manager_test_command(package_manager: Option<&str>) -> String {
     |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ -----------------------------
help: try wrapping the expression in `Some`
     |
1332 |             ProjectType::AstGrepJs => package_manager_test_command(Some(package_manager)),
     |                                                                    +++++               +

error[E0308]: mismatched types
    --> crates/cli/src/commands/init.rs:1334:65
     |
1334 |             ProjectType::Hybrid => package_manager_test_command(package_manager),
     |                                    ---------------------------- ^^^^^^^^^^^^^^^ expected `Option<&str>`, found `&str`
     |                                    |
     |                                    arguments to this function are incorrect
     |
     = note:   expected enum `std::option::Option<&str>`
             found reference `&str`
note: function defined here
    --> crates/cli/src/commands/init.rs:1369:4
     |
1369 | fn package_manager_test_command(package_manager: Option<&str>) -> String {
     |    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ -----------------------------
help: try wrapping the expression in `Some`
     |
1334 |             ProjectType::Hybrid => package_manager_test_command(Some(package_manager)),
     |                                                                 +++++               +

Some errors have detailed explanations: E0308, E0428.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `codemod` (bin "codemod") due to 3 previous errors
```

crates/mcp/src/handlers/package_validation.rs
```diff
@@ -0,0 +1,925 @@
...
+        .filter(|value| !value.is_empty())
+        .map(ToString::to_string);
+
+    PackageJsonLite {
+        scripts,
+        package_manager,
+    }
+}
+
+fn push_missing_file_issues(
```
sahilmob, line 758: Is this required to force agents to generate a valid package? If that's the case, then agents should only use the `init` command and not scaffold manually. Also, this can fail if the project type is `shell`

crates/mcp/src/handlers/package_validation.rs
```diff
@@ -0,0 +1,925 @@
...
+}
+
+fn package_script_command_display(package_manager: &str, script: &str) -> String {
+    match package_manager {
+        "yarn" => format!("{package_manager} {script}"),
+        _ => format!("{package_manager} run {script}"),
+    }
+}
+
+fn infer_package_manager_command(package_root: &Path) -> &'static str {
```
sahilmob: If the validation happens before installing deps. this will return `npm` by mistake

crates/cli/src/commands/init.rs
sahilmob, line 1: I'm assuming that changes in this file are required to prevent agents from package-manager drift, right?

crates/cli/src/templates/ai-native-cli/codemod-cli/commands/codemod.md
```diff
@@ -7,15 +7,25 @@ First classify intent:
 - Otherwise, treat it as codemod discovery or execution: search the registry, pick the best existing package, dry-run it, and apply it only after verification.
 
 Routing:
-- For codemod authoring, call `get_codemod_creation_workflow` and `get_jssg_instructions` from Codemod MCP before proceeding.
-- If the authoring request uses Node/LLRT APIs, capability-gated modules, or non-trivial multi-file JSSG work, also call `get_jssg_runtime_capabilities` from Codemod MCP.
-- If the authoring request implies a monorepo, maintainer workflow, or multi-hop version series, also call `get_codemod_maintainer_monorepo` from Codemod MCP.
-- For codemod discovery or execution, call `get_codemod_cli_instructions` from Codemod MCP for command syntax.
-- When commands fail or produce unexpected behavior, call `get_codemod_troubleshooting` from Codemod MCP.
+- If the expected Codemod MCP tools are not actually available in the callable tool list for this session, stop codemod authoring and tell the user to reload/restart Codex and fix Codemod MCP setup first.
```
sahilmob, line 10: > reload/restart Codex

I think this shouldn't be only specific to Codex

crates/cli/src/templates/ai-native-cli/codemod-cli/commands/codemod.md
```diff
@@ -7,15 +7,25 @@ First classify intent:
...
-- If registry discovery does not yield a suitable package and the user still needs automation, switch to the codemod-authoring path.
-- For codemod authoring, follow the creation workflow guidance from Codemod MCP exactly: stay ast-grep-first, define tests before implementation, keep work inside the requested scope, and do not stop until the package default tests are green.
-- For codemod authoring, do not introduce a shell step just to reach another related file when JSSG can keep both hops inside the same codemod.
-- For codemod execution, follow the CLI instructions from Codemod MCP exactly: dry-run before apply, verify prerequisites, and prefer the current Codemod CLI help and package docs over guesswork.
+- For codemod authoring, inspect only a small representative slice of the repo after or alongside registry discovery, then scaffold and iterate.
+- For codemod authoring, stay AST-first, define fixtures before deep implementation, keep work inside the requested scope, and do not stop until workflow validation, package validation, and the package default tests are green.
+- For codemod authoring, if symbol origin matters, use semantic analysis and binding-aware checks.
```
sahilmob, line 23: Should we point to `docs/jssg/semantic-analysis.mdx` here?

crates/cli/src/templates/ai-native-cli/codemod-cli/commands/codemod.md
sahilmob, line 1: I think this prompt is asking way too much for a single agent to handle. For example, after spending some time implementing a moderately complex codemod, chances are it will forget to run `validate_codemod_package`. Have you tried to split the work between sub-agents https://code.claude.com/docs/en/sub-agents

crates/cli/src/templates/ai-native-cli/codemod-cli/SKILL.md
sahilmob, line 1: Few things are duplicated in `crates/cli/src/templates/ai-native-cli/codemod-cli/commands/codemod.md`

crates/mcp/src/data/prompts/ast-grep-gotchas.md
```diff
@@ -0,0 +1,7 @@
+# ast-grep Hot-Path Gotchas
+
+- Patterns match AST shape, not raw text.
+- Meta-variables capture whole nodes, not partial identifier substrings.
+- Use `dump_ast` when a pattern does not match the expected shape.
+- Do not rename or match identifiers by broad text sweeps alone.
+- If symbol origin matters, use semantic analysis and binding-aware checks instead of text-only matching.
```
sahilmob, line 7: Should we point to docs/jssg/semantic-analysis.mdx here as well?

crates/mcp/src/data/prompts/codemod-creation-workflow.md
sahilmob, line 1: also this one has duplicates

crates/mcp/src/data/prompts/codemod-troubleshooting.md
```diff
@@ -1,60 +1,32 @@
...
 
 Symptom:
-- command aborts because working tree is dirty.
+- the installed `codemod` skill references Codemod MCP tools, but those tools are not in the callable tool list
 
 Fix:
-- review and commit/stash changes, or explicitly allow dirty state:
-  - `codemod workflow run -w my-codemod --target <repo-path> --allow-dirty`
-  - `codemod run <package-name> --target <repo-path> --allow-dirty`
+- reload or restart the Codex session/workspace after `codemod ai` install
```
sahilmob, line 13: > restart the Codex session/workspace after

crates/mcp/src/handlers/package_validation.rs
```diff
@@ -0,0 +1,925 @@
...
+#[derive(Debug, Serialize, schemars::JsonSchema)]
+pub struct ValidateCodemodPackageResponse {
+    pub package_root: String,
+    pub files: PackageFilePresence,
+    pub scripts: PackageScriptSummary,
+    pub workflow_valid: bool,
+    pub test_cases: TestCaseSummary,
+    pub starter_transform_detected: bool,
+    pub generic_readme_detected: bool,
+    pub risky_regex_transform_detected: bool,
```
sahilmob, line 97: I think the above three could be part of `ValidationIssue`

crates/mcp/src/handlers/package_validation.rs
```diff
@@ -0,0 +1,925 @@
...
+    pub command: String,
+    pub success: bool,
+    pub exit_code: Option<i32>,
+    pub timed_out: bool,
+    pub stdout_tail: String,
+    pub stderr_tail: String,
+}
+
+#[derive(Debug, Serialize, schemars::JsonSchema)]
+pub struct ValidationIssue {
```
sahilmob, line 81: I think its better to have a fixed set (enum) of validation issues instead of relying on magic strings

crates/mcp/src/handlers/package_validation.rs
```diff
@@ -0,0 +1,925 @@
...
+    }
+
+    infer_package_manager_command(package_root)
+}
+
+fn detect_package_manager_drift_in_scripts(
+    package_json: &PackageJsonLite,
+    preferred_package_manager: &str,
+) -> Vec<String> {
+    let mismatch_markers = match preferred_package_manager {
```
sahilmob: I think it's better to define it in terms of allowed commands rather than denied commands. This way, we don't have to modify every case whenever we introduce a new package manager.

crates/mcp/src/handlers/package_validation.rs
sahilmob, line 1: I think this would be cleaner as a small typed command model rather than separate execution and display helpers.

Right now `run_package_script()` and `package_script_command_display()` both need to know the package-manager-specific invocation shape (`yarn test` vs `npm run test`), which splits the source of truth. A small `PackageManager` enum plus `PackageCommand::RunScript` would let us derive both argv construction and display formatting from the same value.

For example:

```rust
#[derive(Clone, Copy, Debug, Eq, PartialEq)]
enum PackageManager {
    Npm,
    Yarn,
    Pnpm,
    Bun,
}

impl PackageManager {
    fn binary(self) -> &'static str {
        match self {
            Self::Npm => "npm",
            Self::Yarn => "yarn",
            Self::Pnpm => "pnpm",
            Self::Bun => "bun",
        }
    }
}

enum PackageCommand<'a> {
    RunScript { manager: PackageManager, script: &'a str },
}

impl PackageCommand<'_> {
    fn program(&self) -> &'static str {
        match self {
            Self::RunScript { manager, .. } => manager.binary(),
        }
    }

    fn apply(&self, command: &mut Command) {
        match self {
            Self::RunScript {
                manager: PackageManager::Yarn,
                script,
            } => {
                command.arg(script);
            }
            Self::RunScript { script, .. } => {
                command.args(["run", script]);
            }
        }
    }
}

impl std::fmt::Display for PackageCommand<'_> {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            Self::RunScript {
                manager: PackageManager::Yarn,
                script,
            } => write!(f, "yarn {script}"),
            Self::RunScript { manager, script } => write!(f, "{} run {script}", manager.binary()),
        }
    }
}

sahilmob (changes requested):

sahilmob (approved):

mohebifar (approved):

PR #2080, opened 2026-04-10
chore: simplify init experience

mohab-sameh:
This PR streamlines the `codemod init` flow (defaults + review/edit step) and makes `author` optional in newly scaffolded manifests, with publishing able to infer/normalize author information from authenticated user context.

**Changes:**
- Update `codemod init` to use sensible defaults, add a review screen, and allow omitting `author` in generated `codemod.yaml`.
- Make `CodemodManifest.author` optional and update templates/tests accordingly.
- Enhance `codemod publish` to resolve an “effective” manifest (including inferred `author`) and bundle that effective `codemod.yaml`.

### DIFF
--- a/crates/cli/src/templates/codemod.yaml
+++ b/crates/cli/src/templates/codemod.yaml
@@ -3,7 +3,7 @@ schema_version: "1.0"
 name: "{name}"
 version: "0.1.0"
 description: "{description}"
-{author_block}
+author: "{author}"
 license: "{license}"
 workflow: "workflow.yaml"
 {repository}
--- a/crates/cli/src/templates/codemod.yaml
+++ b/crates/cli/src/templates/codemod.yaml
@@ -3,7 +3,7 @@ schema_version: "1.0"
 name: "{name}"
 version: "0.1.0"
 description: "{description}"
-author: "{author}"
+{author_block}
 license: "{license}"
 workflow: "workflow.yaml"
 {repository}
--- a/crates/cli/src/utils/manifest.rs
+++ b/crates/cli/src/utils/manifest.rs
@@ -1,5 +1,18 @@
 use serde::{Deserialize, Serialize};
 
+pub(crate) const PLACEHOLDER_AUTHOR: &str = "Author <author@example.com>";
+pub(crate) const PLACEHOLDER_AUTHOR_EMAIL: &str = "author@example.com";
+
+pub(crate) fn normalize_author(author: Option<String>) -> Option<String> {
+    author
+        .map(|value| value.trim().to_string())
+        .filter(|value| !value.is_empty())
+}
+
+pub(crate) fn is_placeholder_author(author: Option<&str>) -> bool {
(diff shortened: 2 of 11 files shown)


docs/cli.mdx
```diff
@@ -608,32 +618,32 @@ Both workflows use [trusted publishers](/publishing#trusted-publishers) with OID
...
+
+  Defaults to `Codemod package for <name>`.
 </ResponseField>
 
 <ResponseField name="--author <AUTHOR>" type="string">
   Author name and email.
-  
-  Required in `--no-interactive` mode.
+
+  Defaults to the logged-in user as `username <email>` when available. Otherwise omitted.
```
mohebifar, line 640: Author is required in our backend. Can we just infer the username from `whoami` or something with a placeholder email? During publishing, when we see that placeholder email, we can show a warning to make sure to update the author field.

mohab-sameh: @mohebifar I missed this comment! Good catch that I think this copy is actually inaccurate. What you suggested is the intended behavior. Let me verify again that this is the case and fix either the copy or the behavior.

mohab-sameh: TODO: Enforce placeholder author during scaffolding and warn during publishing.

crates/cli/src/templates/codemod.yaml
```diff
@@ -3,7 +3,7 @@ schema_version: "1.0"
 name: "{name}"
 version: "0.1.0"
 description: "{description}"
-author: "{author}"
+{author_block}
```
mohebifar: author cannot be optional. instead we should try to guess the author name in such cases.

crates/cli/src/utils/manifest.rs
```diff
@@ -6,7 +6,8 @@ pub(crate) struct CodemodManifest {
     pub(crate) name: String,
     pub(crate) version: String,
     pub(crate) description: String,
-    pub(crate) author: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) author: Option<String>,
```
mohebifar: author must be required

crates/cli/src/templates/codemod.yaml
mohebifar: User entry > user name + email if they've logged in > OS username

mohebifar: OS username will be for `--no-interactive` mode only without the `--author`

mohab-sameh: Should we keep it optional in runtime/package validation but enforce it on publishing with fall back to the logged in user?

This allows us to keep the scaffolding experience simple so AI agents do not have to guess an author while creating packages, but simply use the publisher as the author during publishing.

However, @alexbit-codemod mentioned that publsiher != author and they're separate concepts so should we use both interchangeably in publishing flow?

alexbit-codemod:
old closing.

PR #2087, opened 2026-04-15
Fix matrix master completion, workflow status rows, and TUI task visibility

alexbit-codemod:
### What

- **Engine:** Do not mark a matrix **master** complete with **no children** while **`depends_on`** nodes still have non-completed tasks; expose **`get_workflow_state`** for debugging.
- **CLI:** `workflow status` shows **per-matrix-child** id/status (not the master's); add **`--show-state`** for persisted state JSON.
- **TUI:** Task **order** vs `workflow.yaml`, **master visibility** when only the master exists (`from_state`), wider **matrix** columns.

### Risk

Touches **high-impact** paths (workflow correctness, CLI table output, interactive UI). Failures would mostly show as wrong status/order in the UI/CLI; engine mistakes could misrepresent "done." Mitigated by a focused **engine unit test** on the matrix-master + **dependencies not satisfied** case.

### Validation

- `cargo test` (including the new engine test).
- Manual run on **debrrael codemod** — looks good.

### DIFF
--- a/crates/cli/src/tui/task_visibility.rs
+++ b/crates/cli/src/tui/task_visibility.rs
@@ -1,0 +1,84 @@
+//! Which workflow tasks appear in the task list.
+//!
+//! Matrix nodes use a "master" task plus one task per matrix combination. The TUI
+//! historically hid all master tasks to avoid duplicate rows once child tasks exist.
+//! For `strategy.matrix.from_state`, only the master exists until state (e.g. shards)
+//! is written and the scheduler materializes child tasks — hiding the master made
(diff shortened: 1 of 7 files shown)


crates/cli/src/tui/task_visibility.rs
sahilmob, line 1: Do you think we should also hide `install-skill` in the tui?

alexbit-codemod: i'd say that can be addressed in future PRs.

PR #2128, opened 2026-04-23
Workflow TUI foundation

sahilmob:

### DIFF
--- a/crates/core/src/engine.rs
+++ b/crates/core/src/engine.rs
@@ -8,7 +8,6 @@ use std::fs::File;
 use std::future::Future;
 use std::hash::{Hash, Hasher};
 use std::io::Write;
-use std::ops::ControlFlow;
 use std::path::PathBuf;
 use std::process::Stdio;
 use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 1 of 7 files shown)


crates/core/src/engine.rs
```diff
@@ -2534,8 +2478,7 @@ impl Engine {
 
         // Workflows that declare git outputs should use the managed branch/commit/PR path
         // in both cloud and local runs.
-        let manage_git =
-            should_manage_git_for_node(node, self.workflow_run_config.enable_managed_git);
+        let manage_git = should_manage_git_for_node(node);
```
mohebifar, line 2481: This removes the code that I added to force disable managed git in non-tui mode

crates/core/src/engine.rs
```diff
@@ -939,12 +916,7 @@ impl Engine {
                                 .iter()
                                 .find(|node| node.id == task.node_id)
                             {
-                                if engine.workflow_run_config.enable_worktrees
-                                    && should_manage_git_for_node(
-                                        node,
-                                        engine.workflow_run_config.enable_managed_git,
-                                    )
-                                {
+                                if should_manage_git_for_node(node) {
```
mohebifar, line 919: same. this removes disabling git worktree in non-tui mode

mohebifar (changes requested):

crates/core/src/engine.rs
sahilmob: @mohebifar addressed here https://github.com/codemod/codemod/pull/2131

sahilmob:
Closing in favor of reordered replacement PR #2140. The previous stack had PR numbers out of dependency order; the new stack starts at #2140 and preserves the same code via rebuilt branches.

PR #2130, opened 2026-04-23
Fix/manual testing followups

sahilmob:
#### 📚 Description
Disable automatic PR creation and asks for user consent to use git work trees


sahilmob:
Closing in favor of reordered replacement PR #2148. The previous stack had PR numbers out of dependency order; the new stack starts at #2140 and preserves the same code via rebuilt branches.

PR #2131, opened 2026-04-23
Fix/results ignore managed worktrees

sahilmob:
#### 📚 Description

This PR fixes the following:

1. Ignore worktrees from the report
2. Skip worktree and PR creation for non-TUI workflows

#### 🧪 Test Plan

- Should be tested with a non-TUI workflow.
- No regressions for TUI workflows when it comes to PR creation and worktrees


sahilmob:
Closing in favor of reordered replacement PR #2151. The previous stack had PR numbers out of dependency order; the new stack starts at #2140 and preserves the same code via rebuilt branches.

PR #2140, opened 2026-04-24
Workflow TUI foundation

sahilmob:
Replacement PR for the reordered workflow/TUI stack. The old stack had PR numbers out of dependency order; this branch was rebuilt from existing commits and placed in review order.

Includes the non-TUI managed-git/worktree safeguards so this foundation PR is safe and reviewable on its own. Replaces the relevant parts of #2128 and #2131.

### DIFF
--- a/.github/codex/review-output-schema.json
+++ b/.github/codex/review-output-schema.json
@@ -1,0 +1,44 @@
+{
+  "type": "object",
+  "additionalProperties": false,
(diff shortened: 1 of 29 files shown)


mohebifar (approved):

mohebifar:
@sahilmob Looks great to me! I tested it locally. The codex comments all look legit. Can you please look into them? 🙏

PR #2161, opened 2026-04-28
Fix/cloud default pr config

sahilmob:
#### 📚 Description
Enable PR creation on the cloud by default


mohebifar:
Hmm, so it works already. 🤔

mohebifar:
This PR is not actually doing much. And I tested the current CLI version in prod and it worked.

PR #2170, opened 2026-05-04
chore arc codemods findings resolution

sahilmob:
#### 📚 Description
This PR introduces a few enhancements to agents' skills, mainly around using `jssg-utils`. Also introduces new `jssg/utils` based on repeated patterns observed during codemod creation using arc.

Two categories of utils were added
1. Bindings utils
   -`isRuntimeImportBinding`
2. Context utils
   -`unwrapParenthesizedExpression`
   -`isUsedAsConstructor`
   -`isUsedInReflectiveAccess` 

All cases handled by those utilities are documented as unit tests.

Also fixes filesystem path resolution to unlock the utils test cases that require semantic analysis. 

#### 🧪 Test Plan
1. Test suite green.
2. Tested locally on the migrated arc codemods.
3. Utils covered by unit tests


packages/jssg-utils/src/javascript/exports/context.ts
```diff
@@ -0,0 +1,162 @@
...
+  if (
+    (text.startsWith("'") && text.endsWith("'")) ||
+    (text.startsWith('"') && text.endsWith('"')) ||
+    (text.startsWith("`") && text.endsWith("`") && !text.includes("${"))
+  ) {
+    return text.slice(1, -1);
+  }
+
+  return null;
+}
```
mohebifar: I think this code can be improved to use more idiomatic AST-based APIs.

```ts
if (node.is("string")) {
  return node.find({ kind: 'string_fragment' }).text()
}

if (node.is("template_string")) {
  return node.children().slice(1, -1).map(n => n.text()).join("")
}
```

packages/jssg-utils/src/javascript/exports/context.ts
```diff
@@ -0,0 +1,162 @@
...
+  return node
+    ? node
+        .children()
+        .filter((child): child is AnyNode => child.isNamed() && child.kind() !== "comment")
+    : [];
+}
+
+function getFieldNode(node: AnyNode | null, name: string): AnyNode | null {
+  return (node?.field(name as never) as AnyNode | null | undefined) ?? null;
+}
```
mohebifar: Can the types here be improved?

```ts
function getFieldNode<T extends SgNode<...>>(node: T, name: FieldNames<...>): SgNode<...> | null {
 ...
}
```

packages/jssg-utils/src/javascript/exports/bindings.ts
mohebifar, line 1: To me, it looks like this entire file can be simplified. We can just use:

```ts
const def = node.definition({ resolveExternal: false })
```

This keeps the binding lookup scoped to the current module and prevents it from following references into external modules. You can then:

```ts
def.parent().is("whatever") // to check if it's an import statement or function parameter or var declaration or whatever else
```

alexbit-codemod:
whats the status of this? @sahilmob

sahilmob:
@alexbit-codemod Pending approval

PR #2198, opened 2026-05-12
feat(jssg): add xml support

sahilmob:
#### 📚 Description
This PR adds XML support for jssg. ~~Depends on https://github.com/codemod/tree-sitter-parsers/pull/1~~


crates/codemod-sandbox/src/sandbox/engine/codemod_lang.rs
```diff
@@ -18,13 +18,15 @@ use std::str::FromStr;
 pub enum CodemodLang {
     Static(SupportLang),
     Dynamic(DynamicLang),
+    Xml,
```
mohebifar: why is XML a special case in the enum here? a lang is either static or dynamic. we don't want to introduce a new category outside of these two per lang.

mohebifar (commented):
the correct approach will be to build and publish the xml parser in https://github.com/codemod/tree-sitter-parsers/blob/master/languages.json

and similar to less, have it dynamically linked

crates/tree-sitter-loader/src/lib.rs
```diff
@@ -26,54 +27,104 @@ struct DynamicLanguageDefinition {
...
-        extensions: &["less"],
-        expando_char: '_',
-        urls: &[
-            (
-                "macos",
-                "aarch64",
-                concat!(
-                    "https://tree-sitter-parsers.s3.us-east-1.amazonaws.com/tree-sitter/parsers/tree-sitter-less/",
-                    "945f52c94250309073a96bbfbc5bcd57ff2bde49/darwin-arm64.dylib"
+    &[
```
mohebifar, line 75: this bit is getting a tad ugly :D if we could at least make a macro, that'd be cool.

mohebifar (approved):
lgtm! just a small comment.

PR #2267, opened 2026-06-15
docs: clarify shell usage

mohab-sameh:


alexbit-codemod:
wanna incorporate feedback and wrap this? @mohab-sameh

PR #2294, opened 2026-06-30
feat: add agent icons and enhance feedback button UI

siamak:
- Introduced new `agent-icons.tsx` component with `ClaudeCodeIcon` and `CodexIcon`.
- Updated `feedback-button.tsx` to utilize `AgentIcon` for displaying agent options.
- Improved UI elements for loading states and feedback submission confirmation.
- Added constants for activity placeholders and feedback message length limits.
image
image
image
image


AugustinMauroy:
Is it possible to have GH copilot in vscode button ? near 70% of devs use vs-code so IMO it's should be a good thing to allow opening on vs-code. I saw that next.js do it so I think it's possible

alexbit-codemod:
great suggestion @AugustinMauroy. makes sense to me. this PR is for design improvement. we consider your feedback in follow up PRs.

alexbit-codemod (approved):

alexbit-codemod (approved):

PR #2315, opened 2026-07-10
fix(cli): nested codemod capabilities

sahilmob:
#### 📚 Description
This pr fixes the following issue:

Nested codemod steps reused the parent capabilities set unchanged, even though the nested package has its own `codemod.yaml` and its own capabilities.

When the CLI starts a workflow, it still resolves the initial capability set from the top-level codemod, and then the codemod workflow step resolves the child package before executing that child workflow; core reads the child package’s `codemod.yaml`, it parses only the `capabilities` field, and it merges those child capabilities with the inherited parent capabilities. Every step inside the nested workflow receives that merged set.

So if the parent bundle has:

```yaml
capabilities: []
```

```yaml
capabilities:
  - fs
  - child_process
```

the child’s `JSSG` sandbox now receives `fs` and `child_process`, instead of `[]`.


mohebifar:
@sahilmob could you elaborate how this works in the pr description?

mohebifar (approved):

PR #2384, opened 2026-08-12
[CDMD-4896] Add CLI publish tags and SemVer prereleases

sahilmob:
## Summary
- add `codemod publish --tag latest|codemod-builder`, defaulting to `latest`
- send the selected tag in the Registry multipart publish request
- validate manifest versions with standards-compliant SemVer, including Builder prereleases
- resolve Registry `dist_tags` for run/download while preserving exact-version and legacy `latest_version` behavior

## Testing
- `cargo test -p codemod commands::publish::tests` (21 passed)
- `CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo test -p butterflow-core --lib registry::tests` (7 passed)
- `cargo fmt --all -- --check`
- `CARGO_PROFILE_DEV_DEBUG=0 CARGO_PROFILE_TEST_DEBUG=0 CARGO_INCREMENTAL=0 cargo clippy -p codemod -p butterflow-core --tests --no-deps -- -D warnings`

## Manual integration testing
Tested against the local Registry at `http://localhost:3000` with an API token:

- `codemod publish --tag latest` published a prerelease and set only the `latest` tag
- `codemod publish --tag codemod-builder` published `0.0.1-codemod-builder.1` without changing `latest`
- `codemod publish` without `--tag` advanced `latest` while leaving `codemod-builder` pinned
- Registry metadata returned independent `latest` and `codemod-builder` mappings with all published versions present
- default, `@latest`, `@codemod-builder`, and exact prerelease selectors all downloaded and executed successfully through the CLI

## Dependency
Blocked on codemod/codemod-app#949 for Registry dist-tag persistence and API support.

Linear: https://linear.app/codemod/issue/CDMD-4896

### DIFF
--- a/crates/cli/src/commands/publish.rs
+++ b/crates/cli/src/commands/publish.rs
@@ -9,7 +9,7 @@ use anyhow::{anyhow, Result};
 use butterflow_core::utils::validate_workflow;
 use butterflow_core::Workflow;
 use butterflow_models::step::StepAction;
-use clap::Args;
+use clap::{Args, ValueEnum};
 use codemod_llrt_capabilities::module_builder::supported_runtime_external_modules;
 use console::style;
 use log::{debug, info, warn};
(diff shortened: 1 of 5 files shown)


crates/cli/src/commands/publish.rs
```diff
@@ -33,10 +33,30 @@ use crate::commands::TelemetrySenderExt;
...
+
+impl PublishTag {
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::Latest => "latest",
+            Self::CodemodBuilder => "codemod-builder",
+        }
+    }
+}
+
```
mohebifar: Does this have to be an enum? Don't we want to allow any arbitrary side-channels like `beta`, `alpha` etc.?

crates/cli/src/commands/publish.rs
```diff
@@ -601,17 +630,7 @@ fn validate_package_name(name: &str) -> Result<()> {
...
-    if parts.len() != 3 {
-        return false;
-    }
-
-    parts.iter().all(|part| {
-        part.chars().all(|c| c.is_ascii_digit())
-            && !part.is_empty()
-            && (*part == "0" || !part.starts_with('0'))
-    })
+    semver::Version::parse(version).is_ok()
```
mohebifar, line 617: 💯

mohebifar (approved):

PR #2386, opened 2026-08-13
CDMD-4899: add Builder stable publish workflow

sahilmob:
## Summary

- add a CLI-owned command that installs a deterministic Builder publish workflow at the repository root
- scope default-branch runs to the codemod directory
- skip publishing when the package version is unchanged
- preserve the existing tag-based init workflows for external CLI workspaces

Stacked on #2385.

## Validation

- 5/5 workflow-generator tests passed; format and Clippy passed
- verified idempotence, root/nested layouts, per-package workflow names, collision handling, and the existing `init --github-action` flow
- built the Linux ARM64 CLI and ran the generator successfully in a Debian container
- CI note: the failing Rust job is in unrelated flaky `butterflow-core` engine tests

### DIFF
--- a/crates/cli/src/commands/manifest/set_publish_workflow.rs
+++ b/crates/cli/src/commands/manifest/set_publish_workflow.rs
@@ -1,0 +1,258 @@
+use anyhow::{anyhow, bail, Context, Result};
+use clap::{Args, ValueEnum};
+use serde::Serialize;
+use sha2::{Digest, Sha256};
+use std::fs;
+use std::io::Write;
+use std::path::{Path, PathBuf};
+
+use super::set_repository::normalize_repository_directory;
+use crate::commands::publish::validate_package_name;
+
+const WORKFLOW_TEMPLATE: &str = include_str!("../../templates/common/publish-builder.yml");
+const WORKFLOW_DIRECTORY: &str = ".github/workflows";
+
+#[derive(Args, Debug)]
+pub struct Command {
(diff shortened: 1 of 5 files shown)


crates/cli/src/commands/manifest/set_publish_workflow.rs
```diff
@@ -0,0 +1,306 @@
...
+
+use super::set_repository::normalize_repository_directory;
+use crate::commands::publish::validate_package_name;
+
+const WORKFLOW_TEMPLATE: &str = include_str!("../../templates/common/publish-builder.yml");
+const WORKFLOW_DIRECTORY: &str = ".github/workflows";
+static WORKFLOW_PLACEHOLDER: OnceLock<Regex> = OnceLock::new();
+
+#[derive(Args, Debug)]
+pub struct Command {
```
mohebifar, line 19: should this be an `init` subcommand instead?

sahilmob: I considered that, but this command intentionally runs after `init`: Builder has already generated the package and invokes it later when source is connected to an existing repository. Today `init` is a path-taking scaffold command rather than a subcommand namespace, so making this an `init` subcommand would also reinterpret existing CLI invocations. The initial flow still uses `init --github-action`; this command remains the idempotent, non-interactive post-init path.

mohebifar (approved):

PR #2403, opened 2026-08-29
refactor(workspace): bumped to the latest rust version and latest edition.

ppmpreetham:
#### 📚 Description
Bumped to the latest version of Rust, Fixed all the Compatibility issues, Clean Coded the if-let chains, removed refs

#### 🔗 Linked Issue
NIL

#### 🧪 Test Plan
Verify it against the previous test suite with the same changes, it works the same

NIL


crates/cli/src/auth/oidc.rs
```diff
@@ -352,55 +352,52 @@ async fn handle_callback(
...
+                } else if params.contains_key("error") {
+                    let _error = params.get("error").unwrap();
+                    let default_error = "Unknown error".to_string();
+                    let _error_description =
+                        params.get("error_description").unwrap_or(&default_error);
+
+                    let html = format!(
+                        include_str!("html/post-login.html.txt"),
+                        title = "Authentication Failed",
+                        description = "Error: {error}<br>Description: {error_description}"
```
mohebifar: **[P2] Restore interpolation for the callback error details**

`error` and `error_description` are now bound to underscore-prefixed locals, while `description` is a literal `"Error: {error}<br>Description: {error_description}"`. Because that literal is passed as the value of the outer `format!` argument, its braces are not interpolated. OAuth callback failures therefore render the two placeholders rather than the provider error details. Build this value from the locals again (and HTML-escape provider-controlled values before embedding them in HTML).

ppmpreetham: Yup, I've added them back!
Also, I've added the html escape provider which the initial code didn't have either.

sahilmob (changes requested):
Thanks for taking on the Rust 2024 migration. Most of this PR appears to be mechanical compatibility work and formatting.

I have one blocker before approval: the new `unsafe` blocks around `std::env::set_var` and `remove_var` need a real safety guarantee. An `unsafe` block does not make the operation safe—it tells Rust that we verified the conditions ourselves.

In `jssg_execution_service.rs`, `CODEMOD_STEP_ID` is changed inside parallel file-processing code. Multiple threads are already running there, so changing the global process environment is not safe on Unix. Concurrent steps could also read the wrong step ID.

Please:

- Pass the step ID explicitly through the execution/runtime context instead of using a process environment variable.
- Audit the other new environment-related `unsafe` blocks, including tests.
- Add a short `// SAFETY:` explanation only where the required guarantee genuinely holds.
- Add a test covering concurrent JSSG steps with different step IDs and outputs.

The OIDC HTML escaping change makes sense because the error parameters are untrusted browser input. Please add a focused test showing that HTML such as `<script>` is displayed as text and not rendered.

Finally, the branch currently conflicts with `main`, so it should be rebased and CI rerun before approval.

sahilmob:
Hey @ppmpreetham, thank you very much. Could you please rebase this branch so I can review it? 

Sorry for the delay.

ppmpreetham:
No worries, I'll rebase it by tomorrow morning and send ya!

crates/codemod-sandbox/src/sandbox/engine/in_memory_engine.rs
```diff
@@ -125,6 +125,8 @@ pub struct InMemoryExecutionOptions<'a, R> {
     /// will fail to resolve unless the caller separately enables the llrt
     /// `Fs` capability).
     pub fs_sandbox: Option<FsSandbox>,
+    /// Workflow step id recorded with `setStepOutput` calls made by this execution.
+    pub step_id: Option<String>,
```
sahilmob: [P1] The current head does not compile. Adding `step_id` here left three `InMemoryExecutionOptions` initializers incomplete at lines 683, 787, and 837. Both Rust Test and Rust Lint fail with E0063 before tests can run, so the new concurrency coverage has not actually executed in CI.

crates/cli/src/main.rs
```diff
@@ -410,7 +410,10 @@ async fn run_cli() -> Result<()> {
             .map(|params| params.disable_analytics)
             .unwrap_or(false)
     {
-        std::env::set_var("DISABLE_ANALYTICS", "true");
+        // SAFETY: runs on the main thread during CLI startup, before threads spawn.
```
sahilmob: [P1] This safety claim is not true. `run_cli()` is called inside `#[tokio::main]`, which creates a multi-threaded Tokio runtime before this async function runs. This `set_var` therefore executes after worker threads exist; on Unix that violates the operation’s safety requirement and can cause undefined behavior when another thread accesses the environment.

crates/codemod-sandbox/src/sandbox/engine/in_memory_engine.rs
```diff
@@ -1099,7 +1114,10 @@ export default function transform(root) {
             .expect("Failed to write codemod file");
 
         // Seed a host env var that must not leak into process.env.
-        std::env::set_var("PG_SG_SANDBOX_LEAK_CHECK", "should-not-appear");
+        // SAFETY: test is #[serial], so no other test touches the env concurrently.
```
sahilmob: [P2] `#[serial]` only coordinates tests using that serial lock; it does not stop unmarked tests or the Rust test harness from running on other threads. This comment therefore does not prove that no other thread can access the environment while `set_var` runs. The same assumption appears in the other newly added test safety comments.

sahilmob (changes requested):
Thanks — I rechecked the current head. The shared `CODEMOD_STEP_ID` issue is addressed by passing the step ID through the sandbox context, the concurrent coverage is present, and the OIDC escaping test now checks the rendered response. The branch is also mergeable.

The remaining blockers are in the inline findings below. The current Rust Lint and Rust Test jobs both fail on this exact head.

